PR #2, opened 2018-01-23
Changes to allow compilation on macOS

jzuhone:
This PR enables compilation and running on macOS. This only enables compilation using gcc (not clang). 

Things that needed to be changed:

1. Don't include the `malloc.h` header on macOS, whatever is needed in there can be obtained elsewhere.
2. Since we cannot guarantee that a macOS filesystem will be case-sensitive, rename the `GAMER` directory to `Main` since it otherwise conflicts with the `gamer` executable. 
3. Use C++11 `random` instead of `drand` routines, which are not portable. 

TODO:

- [x] Use C++11 `random` in the star formation routines.

@hyschive have a look at these changes and comment before I get into the star formation routines.

Issue #1, opened 2018-01-12
KelvinHelmholtz problem prevents compilation on macOS


jzuhone:
GAMER cannot be compiled on macOS because the KelvinHelmholtz problem uses the thread-safe random number generating functions `drand48_r`, etc., but these are GNU extensions and not portable (they appear to be only on Linux).

I'm looking into what more portable alternatives there may be, but I'm not sure at the moment.

jzuhone:
I've already determined a fix and a couple of other macOS-related fixes. 

I am using gcc--I determined that with clang there are some other compilation bugs. I suppose we should try to address those as well. 

I will have a pull request sometime soon.

hyschive:
Excellent!

Yes, using `clang` will trigger another issue related to the usage of variable-length arrays. I regard it as an important issue but haven't found a portable solution yet.

hyschive:
Fixed with PR #2

PR #3, opened 2018-02-28
More portable random number generator. Addresses Issue #1

hyschive:
This PR adds the header `include/RandomNumber.h` to support different RNG implementations. It currently supports (i) GNU extension `drand48_r` and (ii) C++11 library `<random>`. The purpose is to make the code more portable since GNU extension is not supported on macOS while C++11 is not supported on some old compilers.

To choose different RNG:
1. Set the Makefile option `RANDOM_NUMBER` to either `RNG_GNU_EXT` or `RNG_CPP11`
2. For `RNG_CPP11`, add `-std=c++11` to `CXXFLAG`

@jzuhone Please have a look at it. After it gets merged, we can apply it to your PR #2.

### DIFF
--- a/include/GAMER.h
+++ b/include/GAMER.h
@@ -62,6 +62,7 @@ using namespace std;
 #include "Typedef.h"
 #include "AMR.h"
 #include "Timer.h"
+#include "RandomNumber.h"
 #include "Global.h"
 #include "Prototype.h"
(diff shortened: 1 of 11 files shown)


jzuhone:
Do you just want to merge now? I’d be ok with that.

PR #6, opened 2018-07-27
Support non-equilibrium chemistry in Grackle

hyschive:
This PR implements the non-equilibrium chemistry in Grackle. One can now have `GRACKLE_PRIMORDIAL = 1/2/3` for the 6-, 9-, and 12-species chemistry.

Note that, for now, one still has to manually set `NCOMP_PASSIVE_USER` in the Makefile to match the number of species in the adopted chemistry network.


hisunnytang:
this looks great! it is a lot easier to follow than it was! I would love to see it merged with the base branch.
but we still have to keep in mind that this is not fully self-consistent, as gamma, mean molecular mass is still fixed for the entire simulation box (I wonder how hard it is if we were to implement this, I would love to help). this might be an important trigger for different kinds of instability in first stars (i.e. http://iopscience.iop.org/article/10.1086/507978/meta, Section 5.2, 5.3 )

one minor question is that does yt still read in the species as "Passive00" after this changes, is there a way to fix that?

PR #9, opened 2019-01-02
CPU/GPU integration

hyschive:
This PR integrates the CPU and GPU hydro solvers. It allows for implementing new functionality into GAMER without the need to write CPU and GPU codes separately (in other words, **one doesn't have to learn CUDA** in most cases).

Specifically, all files in `src/Model_Hydro/CPU_Hydro` and `src/Model_Hydro/CPU_HydroGravity` are now linked to `src/Model_Hydro/GPU_Hydro` and `src/Model_Hydro/GPU_HydroGravity`, respectively. These files are thus shared by both pure-CPU and GPU-accelerated simulations.


src/Model_Hydro/CPU_Hydro/CPU_FluidSolver_MHM.cpp
```diff
@@ -297,190 +325,157 @@ void CPU_RiemannPredict_Flux( const real Flu_Array_In[][ FLU_NXT*FLU_NXT*FLU_NXT
...
+   const int  didx_flux[3] = { 1, N_FC_FLUX, SQR(N_FC_FLUX) };
+   const real dt_dh2       = (real)0.5*dt/dh;
+   const real  Gamma_m1    = Gamma - (real)1.0;
+   const real _Gamma_m1    = (real)1.0 / Gamma_m1;
 
-   real dF[3][NCOMP_TOTAL];
-   int ID1, ID2, ID3;
+   const int N_HF_VAR2 = SQR(N_HF_VAR);
+   CGPU_LOOP( idx_out, CUBE(N_HF_VAR) )
+   {
```
zengbs: Could you help me point out the corresponding bracket? Thank you.

hyschive:
> This is an excellent change!
> 
> Just to double check, this version does not support curvi-coordinate, right? ex. in `FullStupUpdate`, there is no information about `Corner`.

That's correct. I'll merge this `cgpu` branch with the curvilinear coordinate extension after this PR is approved.

> 
> And, I must have missed this, could you point me the reason to store `Slope_PPM`, instead of just outputing `FC_Var` in the fluid solver?

`Slope_PPM[]` is still only used by the PPM routine. The only difference is that now it is *allocated only once* during the code initialization so that the same array can be reused for different patches.

PR #11, opened 2019-03-18
Compute radial profiles on the fly

hyschive:
Supported features:
- Comput the profile of a single field on the fly by calling `Aux_ComputeProfile( ... )`
   (an example has been put in the `Example` section of this function) 
- Support both linear and log bins
- MPI and OpenMP parallelization

To-do list:
- [x] Compute multiple fields with a single function call
- [x] User-specified fields (currently only support average density)
- [ ] Other weighting schemes (currently always weighted by cell volume) --> future work
- [ ] Other binning methods (e.g., nth_bin_size=minimum_bin_size*ratio^n) --> future work


hyschive:
@hfhsieh This PR is ready. Could you help review it? Thanks!

include/Macro.h
```diff
@@ -248,21 +248,23 @@
...
+#  define _MAGZ_CC            ( 1L << (NCOMP_TOTAL+ 9) )
+#  define _MAG_ENGY_CC        ( 1L << (NCOMP_TOTAL+10) )
+#  define _DERIVED            ( _VELX | _VELY | _VELZ | _VELR | _PRES | _TEMP | _EINT | _MAGX_CC | _MAGY_CC | _MAGZ_CC | _MAG_ENGY_CC )
+#  define NDERIVE             11
 # else
-#  define _DERIVED            ( _VELX | _VELY | _VELZ | _PRES | _TEMP )
-#  define NDERIVE             5
+#  define _DERIVED            ( _VELX | _VELY | _VELZ | _VELR | _PRES | _TEMP | _EINT )
+#  define NDERIVE             7
 # endif
```
kuochuanpan: We will need to compute the radial profiles of "entropy", "Ye", "Ynu1/2" (neutrino fractions) and "Znu1/2" (neutrino entropies) when doing neutrino transport. 
Perhaps, adding an extra `#ifdef ` (if doing neutrino transport ...) statement.

hyschive: These will be supported automatically as long as the corresponding _XXX symbolic constants are defined in `include/Macro.h`. For the nuclear EoS in your `dev-ccsn` branch, there are already `_ENTR` and `_YE`. So the only thing we need to do is to add the neutrino entropies :)

hyschive: I should clarify that in the comment above I've assumed that `Entr`, `Ye`, and `Znu1/2` should be treated as passive scalars (since we have to evolve them by solving additional continuity equations) instead of derived fields. Right?

kuochuanpan: Yes, `Ye`, `Ynu1`, `Ynu2`, `Znu1`, and `Znu2` are passive scalars. `Entr` could be a derived field.

hyschive: OK. We can add new derived fields in the ccsn branch

hyschive:
@hfhsieh I think it is ready to be merged. Let me know if you have any other suggestions. Thanks!

PR #16, opened 2019-11-21
Enhanced Grackle

wt-liao:
* Allow gamer to set grackle parameters:  three_body_rate, cie_cooling, h2_optical_depth_approximation
* Rescale electron mass: use real electron density in fluid calculation, and rescale it with m_p/m_e when passing it to grackle. It allows normalization of grackle fields.

### DIFF
--- a/include/Global.h
+++ b/include/Global.h
@@ -197,7 +197,7 @@ extern bool            GRACKLE_PE_HEATING;
 extern double          GRACKLE_PE_HEATING_RATE;
 extern char            GRACKLE_CLOUDY_TABLE[MAX_STRING];
 extern int             GRACKLE_THREE_BODY_RATE;
-extern int             GRACKLE_CIE_COOLING;
+extern bool            GRACKLE_CIE_COOLING;
 extern int             GRACKLE_H2_OPA_APPROX;
 extern int             CHE_GPU_NPGROUP;
 #endif
--- a/include/Global.h
+++ b/include/Global.h
@@ -196,6 +196,9 @@ extern bool            GRACKLE_CMB_FLOOR;
 extern bool            GRACKLE_PE_HEATING;
 extern double          GRACKLE_PE_HEATING_RATE;
 extern char            GRACKLE_CLOUDY_TABLE[MAX_STRING];
+extern int             GRACKLE_THREE_BODY_RATE;
+extern int             GRACKLE_CIE_COOLING;
+extern int             GRACKLE_H2_OPA_APPROX;
 extern int             CHE_GPU_NPGROUP;
 #endif
 
--- a/src/Init/Init_Load_Parameter.cpp
+++ b/src/Init/Init_Load_Parameter.cpp
@@ -184,7 +184,7 @@ void Init_Load_Parameter()
    ReadPara->Add( "GRACKLE_PE_HEATING_RATE",    &GRACKLE_PE_HEATING_RATE,         8.5e-26,         0.0,           NoMax_double   );
    ReadPara->Add( "GRACKLE_CLOUDY_TABLE",        GRACKLE_CLOUDY_TABLE,            Useless_str,     Useless_str,   Useless_str    );
    ReadPara->Add( "GRACKLE_THREE_BODY_RATE",    &GRACKLE_THREE_BODY_RATE,         0,               0,             5              );
-   ReadPara->Add( "GRACKLE_CIE_COOLING",        &GRACKLE_CIE_COOLING,             0,               0,             1              );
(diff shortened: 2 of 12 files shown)


hyschive:
@wt-liao Thanks for this contribution and bugfix! I'll take a closer look soon. But it already looks great!

include/Global.h
```diff
@@ -196,6 +196,9 @@ extern bool            GRACKLE_CMB_FLOOR;
 extern bool            GRACKLE_PE_HEATING;
 extern double          GRACKLE_PE_HEATING_RATE;
 extern char            GRACKLE_CLOUDY_TABLE[MAX_STRING];
+extern int             GRACKLE_THREE_BODY_RATE;
+extern int             GRACKLE_CIE_COOLING;
```
hyschive: `int` --> `bool` for `GRACKLE_CIE_COOLING` and `GRACKLE_H2_OPA_APPROX`

src/Main/Main.cpp
```diff
@@ -174,6 +174,9 @@ bool                 GRACKLE_CMB_FLOOR;
 bool                 GRACKLE_PE_HEATING;
 double               GRACKLE_PE_HEATING_RATE;
 char                 GRACKLE_CLOUDY_TABLE[MAX_STRING];
+int                  GRACKLE_THREE_BODY_RATE;
+int                  GRACKLE_CIE_COOLING;
```
hyschive: Declare `GRACKLE_THREE_BODY_RATE` and `GRACKLE_H2_OPA_APPROX` as `bool` instead of `int`, just to be consistent with other Grackle variables.

src/Init/Init_Load_Parameter.cpp
```diff
@@ -183,6 +183,9 @@ void Init_Load_Parameter()
    ReadPara->Add( "GRACKLE_PE_HEATING",         &GRACKLE_PE_HEATING,              false,           Useless_bool,  Useless_bool   );
    ReadPara->Add( "GRACKLE_PE_HEATING_RATE",    &GRACKLE_PE_HEATING_RATE,         8.5e-26,         0.0,           NoMax_double   );
    ReadPara->Add( "GRACKLE_CLOUDY_TABLE",        GRACKLE_CLOUDY_TABLE,            Useless_str,     Useless_str,   Useless_str    );
+   ReadPara->Add( "GRACKLE_THREE_BODY_RATE",    &GRACKLE_THREE_BODY_RATE,         1,               0,             5              );
+   ReadPara->Add( "GRACKLE_CIE_COOLING",        &GRACKLE_CIE_COOLING,             0,               0,             1              );
+   ReadPara->Add( "GRACKLE_H2_OPA_APPROX",      &GRACKLE_H2_OPA_APPROX,           1,               0,             1              );
```
hyschive: To avoid possible confusion, how about adopting the same default values as Grackle (i.e., `GRACKLE_THREE_BODY_RATE=0`, `GRACKLE_CIE_COOLING=0`, and `GRACKLE_H2_OPA_APPROX=0`)?

include/Global.h
wt-liao: Sure. How about set `GRACKLE_CIE_COOLING ` to `bool`, and keep `GRACKLE_H2_OPA_APPROX` in `int` since it is likely to have multiple optical approximations in different version of grackle.

hyschive: Good suggestion

src/Init/Init_Load_Parameter.cpp
```diff
@@ -183,6 +183,9 @@ void Init_Load_Parameter()
    ReadPara->Add( "GRACKLE_PE_HEATING",         &GRACKLE_PE_HEATING,              false,           Useless_bool,  Useless_bool   );
    ReadPara->Add( "GRACKLE_PE_HEATING_RATE",    &GRACKLE_PE_HEATING_RATE,         8.5e-26,         0.0,           NoMax_double   );
    ReadPara->Add( "GRACKLE_CLOUDY_TABLE",        GRACKLE_CLOUDY_TABLE,            Useless_str,     Useless_str,   Useless_str    );
+   ReadPara->Add( "GRACKLE_THREE_BODY_RATE",    &GRACKLE_THREE_BODY_RATE,         0,               0,             5              );
+   ReadPara->Add( "GRACKLE_CIE_COOLING",        &GRACKLE_CIE_COOLING,             0,               0,             1              );
```
hyschive: Change the last two parameters (i.e., min/max range) to `Useless_bool` for a boolean variable

hyschive:
@wt-liao Looks good to me. Finally, can you update the `FormatVersion` variable in `Output/Output_DumpData_Total_HDF5.cpp` and also take note in the `Revision` section?

PR #19, opened 2019-12-10
64-bit bitwise index

### LINKED ISSUE #17: bitwise index approaching limit
the bitwise index (`_deived_field` and `_POTE`) are approaching 32-bit limit when using grackle-12 species.

hyschive:
Declare all bitwise index variables (e.g., `_DENS`) as `long` instead of `int`. Verified by running a Plummer collision test with 40 passive scalars. This resolves #17.
@wt-liao Can you take a look?

### DIFF
--- a/include/Macro.h
+++ b/include/Macro.h
@@ -196,46 +196,46 @@
 #endif
 
 // bitwise field indices
-// --> must have "_VAR_NAME = 1<<VAR_NAME" (e.g., _DENS == 1<<DENS)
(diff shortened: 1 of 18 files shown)


wt-liao:
@hyschive looks great!! 
a minor point: in `tool`, the variables are using `int`. Could caution users on this if they want the analysis on the fly. 
Another thing to keep in mind for future -- some of the deprecated test problems are also using `int`.

hyschive:
@wt-liao Thanks for looking into it! I'll update `tool`.  Let's not worry about those deprecated test problems since they are not compatible with gamer anyway.

PR #30, opened 2020-08-19
Particle ic constructor

CrazyAncestor:
The new test problem is NFW in Testproblem/Hydro.


hyschive (changes requested):
@CrazyAncestor This new functionality is very useful for various projects. Thanks for the great contribution!

I only have two major suggestions.
- [ ] Move most codes currently in `Par_Init_ByFunction_Models.cpp` into a new function (let's call it `Par_SetEquilibriumIC()` for now but you can think of a better name) in `Particle_IC_Constructor.cpp` so that it can be shared by different test problems. It will not only reduce redundant codes but also improve the sustainability and extensibility of this PR.

   The prototype of  `Par_SetEquilibriumIC()` could be, for example, `(Particle_IC_Constructor constructor_Models, real *ParAtt[PAR_NATT_TOTAL], const long ParID0 )`, where `ParAtt[]` is the particle attribute array to be set and `ParID0` is the offset of particle ID (in case there are more than one particle systems to be set; see the next suggestion).
   
- [ ] Allocate an array of `constructor_Models` instead of having multiple particle systems in a single `constructor_Models`. This will make `Particle_IC_Constructor` more concise. Then for the particle system `n` we can just pass `constructor_Models[n]` to `Par_SetEquilibriumIC()`.

   For each `constructor_Models`, set the input runtime parameter filename first (by loading from `Input__TestProb`, for example), and then load **everything else** from this runtime table (including `TypeName` and `Profile_FileName`). You should also add a sanity check to ensure that all requisite parameters are properly provided.
   
Other minor suggestions/comments:
- [x] Does it support units? I guess yes as long as all values in the parameter file are consistent with the units specified by `UNIT_L/M/T`. Do I understand it correctly?
- [ ] Does it support an additional external potential? If yes, provide a concrete example. Does it support both tabular and analytical potential?. If not, we should implement that in a separate PR. This will be very helpful since many astrophysical applications require multiple components (e.g., gas, wave dark matter, black hole).
- [ ] About `Input__TestProb1`:
  - [ ] This filename should be a runtime parameter.
  - [ ] Add a template in `example/input/`.
- [ ] Rename the test problem `NFW` to something more general (e.g., `ParEquilibriumIC`).
- [ ] Remove trailing space.
- [x] Delete `example/test_problem/Hydro/NFW/plot_profile/GAMER_ExtractProfile`.
- [ ] Do your best to follow the code formatting of GAMER, for example,
  - [ ] 3 space indentation
  - [ ] Align codes
  - [ ] Add space and/or blank lines whenever they improve the code readability
- [ ] Add more comments, especially in `Particle_IC_Constructor.cpp`, to help others understand the structure and logic of this new feature.
- [ ] Verify that all plotting and analysis scripts work properly.
- [x] Test OpenMP and MPI parallelization
- [ ] Add a user guide in GAMER wiki.

include/Particle_IC_Constructor.h
```diff
@@ -0,0 +1,120 @@
...
+   int    OPT_EXT_POT;
+   char   ExtPot_Table_Name[MAX_STRING];
+   int      Cloud_RSeed;
+   double   Cloud_Rho0;
+   double   Cloud_R0;
+   double   Cloud_MaxR;
+   double*  Cloud_Center;
+   double*  Cloud_BulkVel;
+   double   Cloud_Einasto_Power_Factor;
+   double   Cloud_Par_Num_Ratio;
```
hyschive, line 40: To make this class more stand-alone, perhaps we should store `Cloud_Par_Num` instead of `Cloud_Par_Num_Ratio` for each cloud. We can still load `Cloud_Par_Num_Ratio` from the cloud runtime table since that's more convenient. But then we immediately convert `Cloud_Par_Num_Ratio` to `Cloud_Par_Num`.

example/test_problem/Hydro/Equilibrium_Cloud/Input__TestProb_Two_Clouds
```diff
@@ -0,0 +1,5 @@
+# problem-specific runtime parameters (mostly filenames of input physical parameters here)
+# The number of Params_Filenames must be the same as Cloud_Num, and please seperate them with spaces.
+
```
hyschive, line 3: We still need `Gas_Background_Dens` here, right?

example/test_problem/Hydro/Equilibrium_Cloud/density_table.txt
```diff
@@ -0,0 +1,1001 @@
+Density Profile
```
hyschive, line 1: Add `#` for header

include/Typedef.h
```diff
@@ -50,6 +50,7 @@ const TestProbID_t
    TESTPROB_HYDRO_MHD_ORSZAG_TANG_VORTEX       =   14,
    TESTPROB_HYDRO_MHD_LINEAR_WAVE              =   15,
    TESTPROB_HYDRO_JEANS_INSTABILITY            =   16,
+   TESTPROB_HYDRO_EQUILIBRIUM_CLOUD             =   17,
```
hyschive, line 53: Code alignment.

src/Particle/Par_EquilibriumIC.cpp
```diff
@@ -0,0 +1,829 @@
...
+    ReadPara->Add( "Cloud_MaxR",         &params.Cloud_MaxR,          0.375,         Eps_double,       NoMax_double      );
+    ReadPara->Add( "Cloud_CenterX",      &params.Cloud_Center[0],     NoDef_double,  NoMin_double,     NoMax_double      );
+    ReadPara->Add( "Cloud_CenterY",      &params.Cloud_Center[1],     NoDef_double,  NoMin_double,     NoMax_double      );
+    ReadPara->Add( "Cloud_CenterZ",      &params.Cloud_Center[2],     NoDef_double,  NoMin_double,     NoMax_double      );
+    ReadPara->Add( "Cloud_BulkVelX",     &params.Cloud_BulkVel[0],    0.0,           NoMin_double,     NoMax_double      );
+    ReadPara->Add( "Cloud_BulkVelY",     &params.Cloud_BulkVel[1],    0.0,           NoMin_double,     NoMax_double      );
+    ReadPara->Add( "Cloud_BulkVelZ",     &params.Cloud_BulkVel[2],    0.0,           NoMin_double,     NoMax_double      );
+    ReadPara->Add( "Cloud_MassProfNBin", &params.Cloud_MassProfNBin,  1000,          2,                NoMax_int         );
+    ReadPara->Add( "Cloud_Par_Num_Ratio", &params.Cloud_Par_Num_Ratio,0.,            0.,                1.0         );
```
hyschive, line 67: After loading `Cloud_Par_Num_Ratio`, convert it to `Cloud_Par_Num`

src/TestProblem/Hydro/Equilibrium_Cloud/Par_Init_ByFunction_Equilibrium_Cloud.cpp
```diff
@@ -0,0 +1,163 @@
...
+         Par_Ratio += Cloud_Constructor.params.Cloud_Par_Num_Ratio;
+         if(k==Filename_Loader.filenames.Cloud_Num-1){
+            Par_Idx_Last = NPar_AllRank;
+         }
+
+         // set equilibrium initial conditions for each cloud
+         // Input Mass_AllRank, Pos_AllRank, Vel_AllRank into this function for particles' masses, positions, and velocities.
+         // NPar_AllRank is the total number (including all clouds) of particles.
+         // Par_Idx_Start is the starting index for the particle in this cloud, and Par_Idx_Last is the last index.
+         Cloud_Constructor.Par_SetEquilibriumIC(Mass_AllRank, Pos_AllRank, Vel_AllRank,NPar_AllRank,Par_Idx_Start,Par_Idx_Last);
```
hyschive, line 97: Revise it according to the simplified parameters of `Par_SetEquilibriumIC()` that I suggest in `Par_EquilibriumIC.cpp`.

hyschive (changes requested):
@CrazyAncestor Good job. My only major suggestion is to simply the parameters of `Par_SetEquilibriumIC()` (see my inline comments).

Some minor comments:
- Check all inline comments carefully.
- Address the remaining points in my previous [review](https://github.com/gamer-project/gamer/pull/30#pullrequestreview-735666939).
- Remove the `x` mode for all new files (use `chmod`).

hyschive (commented):
@CrazyAncestor Just a few remaining tasks.
- [x]  Have you tested the stability of the case with external potential? Can you include such a test in the test problem Equilibrium_Cloud?
- [x] Remove trailing spaces
- [x] Replace tab by space (e.g., by using `retab` with `vim`); remember to realign the codes after that.

You can send me the wiki page later. Let's merge this PR first.

hyschive (approved):

PR #33, opened 2020-09-10
Improving stability

hyschive:
This PR improves the stability of GAMER, especially around low-density valleys.

Major improvement:
1. [INT_OPP_SIGN_0TH_ORDER](https://github.com/gamer-project/gamer/wiki/Runtime-Parameters%3A-Interpolation#INT_OPP_SIGN_0TH_ORDER): Apply 0th-order interpolation if the values to be interpolated change signs in adjacent cells. This prevents from introducing unphysically large velocity in low-density valleys during grid refinement and ghost-cell filling.
2. [OPT__LAST_RESORT_FLOOR](https://github.com/gamer-project/gamer/wiki/Hydro#OPT__LAST_RESORT_FLOOR): When the fluid solver fails and neither 1st-order flux correction ([OPT__1ST_FLUX_CORR](https://github.com/gamer-project/gamer/wiki/Hydro#OPT__1ST_FLUX_CORR)) nor auto dt reduction ([AUTO_REDUCE_DT](https://github.com/gamer-project/gamer/wiki/Runtime-Parameters%3A-Timestep#AUTO_REDUCE_DT)) can help, restore original dt and apply floor values.

Minor improvement:
1. Reduce round-off errors in HLLC when v>>Cs.
2. Revise [INT_MONO_COEFF](https://github.com/gamer-project/gamer/wiki/Runtime-Parameters%3A-Interpolation#INT_MONO_COEFF) by applying only to the smaller slope.
3. Fix a bug in `Flu_FixUp_Flux.cpp` where the array indices were not correctly updated when detecting unphysical results.

@lucyundead and @zengbs, can you help take a look. Thanks!

PR #34, opened 2020-09-11
Isothermal EoS

hyschive:
This PR implements an isothermal equation of state.

**Related options:**
- Compilation options: `EOS=EOS_ISOTHERMAL` and `BAROTROPIC_EOS` (see [here](https://github.com/gamer-project/gamer/wiki/Installation%3A-Simulation-Options#hydro-options))
- Runtime parameters:  `MOLECULAR_WEIGHT` and `ISO_TEMP` (see [here](https://github.com/gamer-project/gamer/wiki/Hydro#ISO_TEMP))

**Warnings:**
- This PR depends on #33. So the latter should be merged first.
- For a barotropic EoS, currently we still *solve* the energy equation but do not *use* the results. Accordingly, the performance is not fully optimized but the physical results should be correct.

@lucyundead Can you help review this PR? Thanks!


lucyundead:
@hyschive I have tested a isothermal run and it gives similar results with other codes. It would be good if the output could also depend on the adopted EoS ( i.e. output pressure for barotropic EoS instead of energy).

hyschive:
@lucyundead Thanks for the feedback! I'll submit another PR to support outputting derived fields like pressure.

PR #41, opened 2020-10-25
[DUP] Special relativistic hydrodynamics

zengbs:
This work is still in progress. See the following list of features:

- [x] VL integrator
- [x] PLM data reconstruction
- [x] PPM data reconstruction
- [x] HLLC Riemann solver
- [x] HLLE Riemann solver
- [x] AMR
- [x] HDF5 output
- [x] OpenMP
- [x] GPU
- [x] MPI
- [x] Bitwise reproducibility
- [x] Self-gravity
- [x] Tabular external gravitational potential
- [x] yt analysis
- [ ] Record the SHA1 of commit in `Record__Note`
- [ ] Wiki [future work]


jzuhone:
Sounds great! Any plans for SR MHD?

zengbs:
Sure. Although the next project seems to have nothing to do with magnetic fields, I plan to develop SR MHD in the near future.

hyschive:
> Sounds great! Any plans for SR MHD?

@jzuhone I think adding SR in MHD should be quite straightforward since the tricky part in SR is the EoS and the conversion between conserved and primitive variables, both of which are independent of B field.

So if there are relevant applications requiring SR-MHD, we will be more than willing to implement it.

jzuhone:
So I started working on a simulation which uses this code. The basic idea is that there is a contact discontinuity between two fluids of dramatically different density (one is the intracluster medium and the other is a relativistic lobe), tilted slightly with respect to the axes of the simulation box, and a jet which comes in from the -y boundary as a boundary condition from within the low-density lobe side and hits the "wall" of intracluster medium. 

For the most part it works as I expect, I just have a couple of questions related to primitive variables:

1. I read the paper and it appears that the primitive variable for the velocity are the three spatial components of the 4-velocity. So, I presume this means that we need to set the components of the velocity in the initial condition using `gamma*vx`, etc, where `vx` is the x-component of the three-velocity. And if I set a user-defined boundary condition, I would have to do the same thing, correct? 

2. What is the code outputting in the data dump files? The spatial components of the 4-velocity u, or the 3-velocity v? I suspect it is the former because the components are sometimes greater than 1 in units of c.

jzuhone:
Ah, I had forgotten that the code dumps out momentum densities instead. So these then are the M_j from the paper?

hyschive:
@zengbs Please help clarify the above questions. Also, could you share the yt routines (or even better a working yt branch) to compute primitive variables?

@jzuhone I will add a user interface soon to output user-specified fields (so one can store additional primitive variables in snapshots if needed).

zengbs:
@jzuhone 

- As you would expect, the primitive variables for velocity in sr-hydro solver are defined by three spatial components of four-velocity. The initial conditions (run-time parameters) for velocity are described by three spatial component of four-velocity.

- If I understand correctly, we cannot set up the user-defined B.C. only through run-time parameters but `BC_User_Template()`. In doing so, we can either set user-defined B.C. by conserved variables (`fluid[]`) or by primitive ones with the help of `Hydro_Pri2Con()`.

- When `MODEL=SRHD`, HDF5 files store mass density (Eq. 5a), momentum densities (Eq. 5b), reduced energy density (Eq. 6), and `kT/mc^2` (Eq. 16). Dumping `kT/mc^2` is to avoid root-finding during yt visualization. (Here Eqs are listed in my paper)

- The plot script for SRHD have not been merged with yt yet, but you can clone the script. This script is flexible enough to design the layout of figures and to change plotting parameters easily (e.g. the limitation of colorbar, the coordinate where slice passing...etc.) only by loading pure **text** files. Thus, making visualization is free from the knowledge of yt or python.

1. `cd WORKING_DIRECTORY`
2. `git clone https://github.com/zengbs/plot_scripts.git`
3. `cp $WORKING_DIRECTORY/plot_scripts/Plot__Parameters/plot__MultipleSlice .`
4. Modify the text file `plot__MultipleSlice` to change plotting parameters. (e.g. layout, the limitation of colorbar, the coordinate where slice passing etc.)
5. `python $WORKING_DIRECTORY/plot_scripts/Read__Parameters/general__ReadSliceMultiple.py -f plot__MultipleSlice`
6. Complete visualization. e.g.,
 slice_00
 
 Please note that this script currently supports 2D off-axis slices, 2D axis-aligned-slices, and 1D ray plots.

hyschive:
> @zengbs @hyschive thanks, all of this makes a lot of sense.
> 
> I'd be happy to make the required changes in yt and submit a PR if that helps--we'd only need to change the field definitions for when the simulation includes relativity. We could detect a flag for SRHD in the output file when loading the dataset, which would then trigger the new field definitions. If either of you have a preferred flag for this let me know, I saw some when I was inspecting the file yesterday.

@jzuhone That will be super helpful. Thanks! We can use this [flag](https://github.com/zengbs/gamer/blob/to-be-merged/src/Output/Output_DumpData_Total_HDF5.cpp#L2592).

jzuhone:
Hi guys, 

So it turns out I am having a bit of trouble. 

As I mentioned above, I have a user boundary condition which is outflow except for a small circle, at which there is a constant inflow of material. 

@hyschive unfortunately to get this to work I had to edit the signature of `Flu_BoundaryCondition_User` and the `BC_User_Ptr` because the easiest way to get a boundary condition like this is for the `BC_User_Ptr` to know about the cell structure just like the built-in hydro outflow boundary condition does, so I pass it the cell index and the `Array` pointer, as well as some other information. Only within the circle where the jet is inflowing do I set the jet condition. 

But I am getting error messages such as:

`E^2+2*E*D-|M|^2= 0.0000000e+00`

And it turns out that I get these error messages even if I shut the jet itself off but keep the custom boundary condition (so what it should do is outflow everywhere). 

You can see my edits here:

https://github.com/jzuhone/gamer/blob/jet_icm_wall/src/Fluid/Flu_BoundaryCondition_User.cpp

And my problem set up with its custom BC here:

https://github.com/jzuhone/gamer/blob/233f26c41206dc4cf47f8e8a9bab6e2a183cce6d/src/TestProblem/Hydro/JetICMWall/Init_TestProb_Hydro_JetICMWall.cpp#L253

Any idea why it may not be working? Sorry to bug you guys. Otherwise my setup is working great.

zengbs:
@jzuhone Thanks for your report! I would like to run your code to identify the problem. Push your newest code to your repository and tell me the branch name, please.

jzuhone:
Thanks so much, @zengbs!

My branch is here:

https://github.com/jzuhone/gamer/tree/jet_icm_wall

which is branched off of the branch on this PR.

You can get a tar.gz of the `Input__` files and the `Makefile` by doing this:

`curl -JO http://use.yt/upload/023a1708`

You'll have to tweak the libraries in the Makefile, and maybe the grid size, etc in the parameter files.

hyschive:
@jzuhone Thanks. We will look into it.

zengbs:
@jzuhone I follow your suggestion to set up the boundary condition. i.e. put a circular jet source on the boundary. You can see the newest code and compare with the old one:

`git clone https://github.com/zengbs/gamer.git`
`git checkout --track origin/jet_icm_wall`
`git diff 233f26c4 773784cb`

In addition to the above thing, we recommend the following suggestions when SRHD is enabled:

- Enable the macro BITWISE_REPRODUCIBILITY in Makefile.
- Use the criterion (1+M^2)*epsilon (Eq.20) to decide whether to use single or double precision, where M is the Mach number and epsilon is the machine epsilon.
- Use the spatial component of 4-velocity as input parameter; otherwise we will be forced to use gamma=1.0/sqrt(1.0-Velocity*Velocity), which will suffer from catastrophic cancellation when Velocity approach 1.

 Feel free to let me know if you have more questions!

Input__*, plot__MultipleSlice, and Makefile are [here](https://github.com/gamer-project/gamer/files/5860268/jzuhone.tar.gz).
fig
From left to right: The slices passing x=0.5, z=0.5, and y=0.0.

zengbs:
@jzuhone Thanks for the wait!
There are two bugs in the version `233f26c4` (you originally provided):

- If `NVar_Flu != NCOMP_FLUID` at this line:
https://github.com/jzuhone/gamer/blob/233f26c41206dc4cf47f8e8a9bab6e2a183cce6d/src/TestProblem/Hydro/JetICMWall/Init_TestProb_Hydro_JetICMWall.cpp#L284
then this line fails:
https://github.com/jzuhone/gamer/blob/233f26c41206dc4cf47f8e8a9bab6e2a183cce6d/src/Fluid/Flu_BoundaryCondition_User.cpp#L215
Because `SRHD_Con2HTilde()` needs `BVal[]`  that have `NCOMP_TOTAL` fields.
- The section:
https://github.com/jzuhone/gamer/blob/233f26c41206dc4cf47f8e8a9bab6e2a183cce6d/src/Fluid/Flu_BoundaryCondition_User.cpp#L215-L220

Hence, we provide a temporary solution to meet your requirement, please:
`git clone git@github.com:zengbs/gamer.git`
`git checkout jet_icm_wall_zengbs`

Note that this temporary solution does not support two refinement criteria `OPT__FLAG_LOHNER_PRES` and `OPT__FLAG_LOHNER_TEMP` but supports others. If you would like to resolve patches based on pressure, please use `OPT__FLAG_PRES_GRADIENT` instead. This is because (1) refining patches based on the pressure gradient does not use the information on the boundary but the Lohner pressure does. (2) the line below will fail when we prepare the patch composed of pressure or temperature for Lohner criterion. i.e. `NVar_Flu=0` will trigger crash.
https://github.com/zengbs/gamer/blob/2ab386d6a5d3e3e89bde097a1ddb1b19ea96a21e/src/TestProblem/Hydro/JetICMWall/Init_TestProb_Hydro_JetICMWall.cpp#L291

In addition to the above things, the following items also need extra attention:
- The function `BC()` should return `NCOMP_FLUID` fields, for both `SRHD` and `HD`.
- Before `SRHD` simulation, use the criterion `(1+M^2)*epsilon` (Eq.20) to decide whether to adopt single or double precision, where `M` is the Mach number and `epsilon` is the machine epsilon. i.e. Calculate Eq(20) by hand before simulation.
- Use the spatial component of 4-velocity, instead of 3-velocity, as input parameter; otherwise we will be forced to use the equation `gamma=1.0/sqrt(1.0-Velocity*Velocity)`, which will suffer from catastrophic cancellation when Velocity approach 1. See the paragraph above Eq(8) in my paper :).

Feel free to let me know if you have more questions!

Input__*, plot__MultipleSlice, and Makefile are [here](https://github.com/gamer-project/gamer/files/5865339/jzuhone.tar.gz).

hyschive:
@jzuhone As @zengbs said, the major issue in your implementation of mixed outflow and inflow BC's is that the user-defined routine `BC()` should always return `NCOMP_TOTAL` fields (i.e., all `NCOMP_FLUID` major fluid variables plus passive scalars if any), as mentioned [here](https://github.com/jzuhone/gamer/blob/jet_icm_wall/src/Fluid/Flu_BoundaryCondition_User.cpp#L27).

I think the following hard-coded solution can support options like `OPT__FLAG_LOHNER_PRES` and `OPT_FLAG_LOHNER_TEMP`, although not in an elegant way:

1. Pass both `NVarCC_Flu` and `NVarCC_Der` to `Flu_BoundaryCondition_User()` [here](https://github.com/jzuhone/gamer/blob/jet_icm_wall/src/Main/Prepare_PatchData.cpp#L1666) (and other places calling `Flu_BoundaryCondition_User()` as well).
2. Then pass `NVar_Flu + NVar_Der` instead of `NVar_Flu` to `BC_User_Ptr()` [here](https://github.com/jzuhone/gamer/blob/jet_icm_wall/src/Fluid/Flu_BoundaryCondition_User.cpp#L178). Let's call it `NVar_All`.
3. Replace [your outflow BC](https://github.com/jzuhone/gamer/blob/jet_icm_wall/src/TestProblem/Hydro/JetICMWall/Init_TestProb_Hydro_JetICMWall.cpp#L283) by the following lines:
```
        for (int v=0; v<NVar_All; v++)
            Array3D[v][idx[2]][idx[1]][idx[0]] = Array3D[v][idx[2]][j_ref][idx[0]];
```
4. Here is the slightly trick part. [These lines](https://github.com/jzuhone/gamer/blob/jet_icm_wall/src/Fluid/Flu_BoundaryCondition_User.cpp#L181-L248) should only be applied to the cells with the *inflow* BC but NOT those with the *outflow* BC. It is because the outflow BC (including all derived fields like pressure and temperature) have been applied by the step 3 above.

I haven't tested it but you may give it a try. Feel free to let us know if there are any further problems.

jzuhone:
@hyschive @zengbs thanks for all of this effort, it's much appreciated. I will check this out. 

I am not wedded to the Lohner refinement criteria in this case--what we want is any refinement criteria which will refine on jumps in density at the very least.

src/Fluid/Flu_BoundaryCondition_User.cpp
```diff
@@ -200,9 +203,20 @@ void Flu_BoundaryCondition_User( real *Array, const int NVar_Flu, const int Arra
...
+      real LorentzFactor, Prim[NCOMP_FLUID];
+
+      Hydro_Con2Pri( BVal, Prim, (real)NULL_REAL, NULL_BOOL, NULL_INT, NULL, NULL_BOOL,
+                     (real)NULL_REAL, EoS_DensEint2Pres_CPUPtr, EoS_DensPres2Eint_CPUPtr,
+                     EoS_GuessHTilde_CPUPtr, EoS_HTilde2Temp_CPUPtr, EoS_AuxArray_Flt,
+                     EoS_AuxArray_Int, h_EoS_Table, NULL, &LorentzFactor );
+
+      if ( PrepLrtz )   Array3D[ v2 ++ ][k][j][i] = LorentzFactor;
+#     else
       if ( PrepVx   )   Array3D[ v2 ++ ][k][j][i] = BVal[MOMX] / BVal[DENS];
```
jzuhone, line 216: @zengbs just a quick question--are lines 216-218 which prepare the velocity still valid in SR?

jzuhone: Or I guess they are not even called if `SRHD` is defined. So we don't need to prepare the velocities here?

hyschive: @jzuhone Actually, these lines of preparing velocities are currently not used in either HD or SRHD. Only momenta are used.

jzuhone:
@zengbs sorry, one other question--why do we need this line?

https://github.com/zengbs/gamer/blob/2ab386d6a5d3e3e89bde097a1ddb1b19ea96a21e/src/TestProblem/Hydro/JetICMWall/Init_TestProb_Hydro_JetICMWall.cpp#L282

hyschive:
Hi John,
I think that line is irrelevant.

Hi Po-Hsun,
Can you remove it?

Cheers,
Hsi-Yu

zengbs:
@jzuhone  That line is for debug only and have been removed. :)

hyschive:
@zengbs I assume this PR has been replaced by #141. Should we close it?

@zhanguihan is still busy but will start to review this PR once finishing his current project.

jzuhone:
@hyschive I presume you can close this now.

hyschive:
Duplicate of #185

PR #47, opened 2021-01-20
Gamer cmz

lucyundead:
Add CMZ test problem
Extended read-in potential table to have different dh along three directions


hyschive (approved):
@lucyundead Looks good. Thanks for the contribution!

There are some minor compilation issues that I'll fix immediately in the next commit.

PR #49, opened 2021-02-02
Change the default value of GFUNC_COEFF0

JeiwHuang:


hyschive (approved):
@JeiwHuang Thanks for the great efforts you put in!

PR #51, opened 2021-02-02
Update the job script for Taiwania2

JeiwHuang:


example/queue/submit_taiwania2.job
```diff
@@ -1,21 +1,21 @@
...
+#SBATCH --time=02:00:00                   ## time limit (hh:mm:ss)
+#SBATCH --output=log-%j                   ## stdout output file
 
-export OMP_PROC_BIND="spread"	## make cores in order inside a task
+#Load module
+module purge
+module load compiler/gnu/4.8.5 nvidia/cuda/10.0 openmpi/3.1.4
 
-srun --cpu-bind=v,cores -m block:block ./gamer
+export OMP_PROC_BIND="spread"  ## make cores in order inside a task
```
hyschive, line 19: If this line still necessary?

example/queue/submit_taiwania2.job
```diff
@@ -1,21 +1,21 @@
...
 
-##srun --hint=nomultithread ./gamer
-##srun ./gamer
+srun --cpu-bind=v,cores -m block:block ./gamer 1>>log 2>&1
```
hyschive, line 21: Is it still necessary to add `--cpu-bind=v,cores -m block:block`?

example/queue/submit_taiwania2.job
```diff
@@ -1,21 +1,21 @@
...
-module load nvidia/cuda/10.1 compiler/intel/2018 openmpi3/3.1.4
+#SBATCH --job-name=YOUR_JOB_NAME          ## job name
+#SBATCH --mail-type=ALL                   ## mail notification type
+#SBATCH --mail-user=YOUR_EMAIL            ## mail address
+#SBATCH --account=YOUR_ACCOUNT            ## account (https://www.twcc.ai/user/dashboard)
+#SBATCH --partition=gp1d                  ## queue name (gtest/gp1d/gp2d/gp4d)
+#SBATCH --nodes=2                         ## number of nodes
+#SBATCH --gres=gpu:8                      ## number of GPUs per node
+#SBATCH --ntasks-per-node=8               ## number of tasks per node
+#SBATCH --cpus-per-task=4                 ## number of CPUs per task
```
hyschive, line 11: 1. Does using 8 instead 4 threads per MPI process improve the performance (i.e., `--cpus-per-task=8` and `OMP_NTHREAD=8`)?
2. Does using multiple MPI processes to access the same GPU improve the performance? For example,
```
#SBATCH --gres=gpu:8                 ## number of GPUs per node
#SBATCH --ntasks-per-node=16         ## number of tasks per node
#SBATCH --cpus-per-task=2            ## number of CPUs per task
```
For the above example, one will also need to set `OMP_NTHREAD = 2` in `Input__Parameter`.

example/queue/submit_taiwania2.job
JeiwHuang: It can make the cpu core ID in the same order as the OpenMP thread ID. But this is not necessary.

example/queue/submit_taiwania2.job
JeiwHuang: It is not necessary.

example/queue/submit_taiwania2.job
JeiwHuang: 1. There is a limit that the number of CPUs per GPU cannot be more than 4. (1 GPU : 4 CPU)
2. No, this is slightly slower.

hyschive (approved):
@JeiwHuang Looks good. Thanks for the contribution!

PR #55, opened 2021-05-12
Add function to record git SHA1 hash to Record_Note

xuanweishan:
By editing Makefile, setting a variable GIT_SHA1='git rev-parse HEAD' to get git hash.

### DIFF
--- a/include/Macro.h
+++ b/include/Macro.h
@@ -883,6 +883,9 @@
 #  define ATAN2( a, b )    atan2f( a, b )
 #endif
 
+// stringizing operator in C/C++
+#define STR( x )            #x
+#define SHOW_MACRO( x )     STR( x )
 
 // sign function
 #define SIGN( a )       (  ( (a) < (real)0.0 ) ? (real)-1.0 : (real)+1.0  )
--- a/include/Macro.h
+++ b/include/Macro.h
@@ -883,10 +883,6 @@
 #  define ATAN2( a, b )    atan2f( a, b )
 #endif
 
-// stringizing operator in C/C++
-#define STR( x )            #x
-#define SHOW_MACRO( x )     STR( x )
-
 // sign function
 #define SIGN( a )       (  ( (a) < (real)0.0 ) ? (real)-1.0 : (real)+1.0  )
 
--- a/src/Auxiliary/Aux_TakeNote.cpp
+++ b/src/Auxiliary/Aux_TakeNote.cpp
@@ -1460,6 +1460,13 @@ void Aux_TakeNote()
       fprintf( Note, "***********************************************************************************\n" );
       fprintf( Note, "\n\n" );
 
+//    record the branch name in git
+      fprintf( Note, "branch name in git\n" );
+      fprintf( Note, "***********************************************************************************\n" );
+      fprintf( Note, "%s\n", SHOW_MACRO( GIT_BRANCH ));
+      fprintf( Note, "***********************************************************************************\n" );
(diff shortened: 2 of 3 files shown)


include/Macro.h
```diff
@@ -883,6 +883,9 @@
 #  define ATAN2( a, b )    atan2f( a, b )
 #endif
 
+// stringizing operator in C/C++
+#define STR( x )            #x
+#define SHOW_MACRO( x )     STR( x )
```
hyschive, line 888: Use the existing macros `QUOTE` and `EXPAND_AND_QUOTE` defined [here](https://github.com/gamer-project/gamer/blob/master/include/Macro.h#L915-L916).

src/Makefile
```diff
@@ -231,6 +231,7 @@ SIMU_OPTION += -DRANDOM_NUMBER=RNG_GNU_EXT
 ##CXXFLAG     = -g -fast
 # CXXFLAG    += -w1                                       # warning flags
 # OPENMPFLAG  = -fopenmp                                  # openmp flag
+# CXXFLAG    += -DGIT_SHA1="`git rev-parse HEAD`"         # record SHA1 in git
```
hyschive, line 234: - Also record the git branch by `git rev-parse --abbrev-ref HEAD`.
- Put this line somewhere else since it should be independent of the adopted compiler (e.g., below [here](https://github.com/gamer-project/gamer/blob/master/src/Makefile#L780)).
- Print a warning message if `git` doesn't work (instead of aborting the compilation)

src/Makefile
```diff
@@ -258,6 +259,7 @@ SIMU_OPTION += -DRANDOM_NUMBER=RNG_GNU_EXT
                -Wno-maybe-uninitialized -Wno-unused-but-set-variable \
                -Wno-unused-result -Wno-unused-function
  OPENMPFLAG  = -fopenmp                                  # openmp flag
+ CXXFLAG    += -DGIT_SHA1="`git rev-parse HEAD`"         # record SHA1 in git
```
hyschive, line 262: Remove this redundant line.

src/Auxiliary/Aux_TakeNote.cpp
```diff
@@ -1453,6 +1453,12 @@ void Aux_TakeNote()
       fprintf( Note, "***********************************************************************************\n" );
       fprintf( Note, "\n\n");
 
+//    record the SHA1 hash in git
+      fprintf( Note, "SHA1 in git\n" );
+      fprintf( Note, "***********************************************************************************\n" );
+      fprintf( Note, "%s\n", SHOW_MACRO( GIT_SHA1 ));
```
hyschive, line 1459: Record the git branch too. See the comment below about how to do that.

hyschive (changes requested):
@xuanweishan Looks good. Thanks for the contribution! I've added some minor suggestions.

src/Makefile
hyschive: @xuanweishan 
- Since this is not really a compiler flag, I would suggest storing it in a separate variable (e.g., `GIT_INFO`) and adding it [here](https://github.com/gamer-project/gamer/blob/master/src/Makefile#L931). Also, remove the extra whitespace below [here](https://github.com/gamer-project/gamer/blob/master/src/Makefile#L884).
- If `git` returns any error, set both `GIT_SHA1` and `GIT_BRANCH` to `"Unknown"`.

hyschive (approved):

PR #56, opened 2021-05-25
Perform YT_Inline() at initialization

cindytsai:


hyschive (changes requested):
@cindytsai There is no need to modify the `Makefile` in this PR, right?

cindytsai:
No, this version is not paralleled yet.
I accidentally make changes to `Makefile` when merging.
Should I submit a new one?

hyschive:
@cindytsai You can just restore the original `Makefile` and then push it to the same branch. There is no need to submit a new PR.

hyschive (approved):

Issue #71, opened 2021-08-08
Truncating cluster gas profile


hyschive:
The current initial condition of the `ClusterMerger` test problem does not apply truncation in gas profiles. As a result, the ambient gas mass becomes highly non-negligible (it’s comparable to the cluster virial mass), which eventually leads to shock stalling. 

https://user-images.githubusercontent.com/18057331/128619966-c2207915-1d53-4915-91e5-a6e472e3cec4.mp4

PR #82, opened 2021-09-23
Update ICs for ClusterMerger problem and change behavior of profiles near boundaries

jzuhone:
Recently it was pointed out that the merger shocks in the ClusterMerger problem stall because the gas profiles are not truncated (see issue #71). This PR supplies new profiles which fall off faster, as well as provide the script which is used to create them using the https://github.com/jzuhone/cluster_generator package. Unfortunately, this package is still under development and does not yet have good documentation, but most functions have docstrings. 

I have also changed the way that the edges of the profiles are handled--outside of the maximum radii of the cluster profiles, the physical quantites are constant and the velocities are zero (this assumes that density, temperature, and pressure are very small at large enough radius). 

I also made the box size somewhat bigger, because for the existing test problem the magnetic field at the boundary would flow toward the clusters, aided by the velocity of the clusters near the edge, and stretch the field to very large strengths, causing instability.


hyschive:
@jzuhone Thanks for the update! I'll review it soon.

hyschive (approved):
@jzuhone Looks good. Thanks for updating this setup and sharing the IC script!

PR #94, opened 2022-01-04
Better handling for variable-length array pointers

jzuhone:
This PR changes a number of places throughout the code where a 1D array was pointed to by a 3D array using array sizes not defined at compile time. This is ok with compilers like gcc but not with clang. 

@hyschive I may have missed some other places where we do this. Also, could you test yourself?

### DIFF
--- a/src/Fluid/Flu_BoundaryCondition_User.cpp
+++ b/src/Fluid/Flu_BoundaryCondition_User.cpp
@@ -158,7 +158,8 @@ void Flu_BoundaryCondition_User( real *Array, const int NVar_Flu, const int Arra
 
 
 // 1D array -> 3D array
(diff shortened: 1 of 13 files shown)


hyschive:
@jzuhone Thanks! Here are the remaining files requiring this fix:
- `SelfGravity/Poi_BoundaryCondition_Extrapolation.cpp`
- `SelfGravity/CPU_Poisson/CPU_PoissonSolver_MG.cpp`
- `Model_Hydro/MHD_BoundaryCondition_Outflow.cpp`
- `Model_Hydro/MHD_BoundaryCondition_User.cpp`
- `Model_Hydro/MHD_BoundaryCondition_Reflecting.cpp`
- Also this [line](https://github.com/gamer-project/gamer/blob/master/src/Particle/Par_Aux_GetConservedQuantity.cpp#L150)

PR #96, opened 2022-01-25
Diode boundary condition

jzuhone:
This PR implements a "diode" boundary condition, which is identical to "outflow" except that vector quantities are only able to "flow out", i.e. velocity and magnetic field vectors normal to the boundary can point outward but not inward.

https://flash.rochester.edu/site/flashcode/user_support/flash4_ug_4p62/node57.html#SECTION05141000000000000000

I'm still trying to think of a good test problem for this.


jzuhone:
I actually just realized that I should't be doing this for the magnetic fields--will fix.

hyschive:
@jzuhone Is it ready for review?

jzuhone:
@hyschive yes

jzuhone:
@hyschive when you have time, could you have a look at this one?

hyschive:
@jzuhone Sorry for the belated action. This looks straightforward and I'll review/merge it shortly. Just two quick questions.
- Why doesn't it support MHD?
- Any recommended test for it?

jzuhone:
Hi @hyschive, 

1. I made a mistake--it does support MHD. That's fixed now. 
2. So ideally the `ClusterMerger` problem should have diode BC by default. I've set that by default now in its `Input__Parameter`. I'm going to have a postdoc check it for her simulations. Is there any other test you would like?

hyschive:
@jzuhone Thanks for the update. I will merge it once it is confirmed to work properly in your `ClusterMerger` simulations.

src/Model_Hydro/Hydro_BoundaryCondition_Diode.cpp
```diff
@@ -0,0 +1,618 @@
...
+// set the boundary values
+   for (int v=0; v<NVar_Flu; v++)
+   {
+      TFluVarIdx = TFluVarIdxList[v];
+
+      if ( TFluVarIdx == MOMX )
+      {
+         for (int k=Idx_Start[2]; k<=Idx_End[2]; k++)    {
+         for (int j=Idx_Start[1]; j<=Idx_End[1]; j++)    {
+         for (int i=Idx_Start[0]; i<=Idx_End[0]; i++)    {  ii = i_ref - i;
```
hyschive, line 153: The definitions of `j/k_ref` and `ii/jj/kk` are incorrect. In short, they should follow the same definitions in `Hydro_BoundaryCondition_Outflow.cpp` instead of `Hydro_BoundaryCondition_Reflecting.cpp`. I've put a correct version at [here](https://gist.github.com/hyschive/96273166d984851778bc8ff6976f44cf).

hyschive (changes requested):
@jzuhone Thanks for the contribution! I have a closer look at this PR. In addition to the major comment about `Hydro_BoundaryCondition_Diode.cpp`, please also increase the upper limits of `OPT__BC_FLU_*` in [Init_Load_Parameter()](https://github.com/jzuhone/gamer/blob/diode/src/Init/Init_Load_Parameter.cpp#L60-L65) to `5`.

I've been testing the corrected code using the Plummer test problem and it seems to work well. I'll post the results soon. But note that the test doesn't include MHD.

hyschive:
@jzuhone Results of Plummer sphere with an explosion triggered at `t~3`.

Slice plot using the diode BC.

https://user-images.githubusercontent.com/18057331/184667629-a7c298a5-7bd3-4064-8248-84d201c0164d.mp4

The density evolution along the x-axis through the Plummer center. For the outflow BC, the density in the outskirts increases before the explosion since it allows gas to flow "inward". In comparison, for the diode BC, the density in the outskirts decreases since inflow is prohibited.

https://user-images.githubusercontent.com/18057331/184667657-e1b943a8-9cf9-4749-bf30-90b067df8a56.mp4

Total mass as a function of time. It further confirms that, before the explosion, the total mass with the outflow BC increases while that with the diode BC is roughly constant.

fig_mass

If you agree, I'll merge this PR once the aforementioned two issues are resolved.

hyschive (approved):

jzuhone:
I was about to post a movie showing that it works for MHD--but never mind!

PR #115, opened 2022-04-30
Add yt script for compute particle number density

koarakawaii:
1. Add `yt` derived field `number density`, using field `("deposited", "all count")` divided by field `("index", "cell volume")`
2. Compute number density profile using the defined derived field
3. Save the profile as text file via `numpy`


hyschive (approved):
@koarakawaii Looks good. Thanks for the contribution!

PR #127, opened 2022-06-27
 [Bugfix] File name change density_table

BarryTC:
The parameter Density_Table_Name only accepts file names without explicit file format. This update fixes the format definition of the template density_table.


hyschive:
Addressed by c57da444f6ac0be977db18fe6817f3f20110bafc

PR #134, opened 2022-08-03
User-specified operations in EvolveLevel()

hyschive:
Add the following two user-specified function pointers in `EvolveLevel()`
- `Mis_UserWorkBeforeNextLevel_Ptr`
- `Mis_UserWorkBeforeNextSubstep_Ptr`

### DIFF
--- a/src/Main/EvolveLevel.cpp
+++ b/src/Main/EvolveLevel.cpp
@@ -768,7 +768,7 @@ void EvolveLevel( const int lv, const double dTime_FaLv )
 // ===============================================================================================
 
 
-//    14. user-specified operations before proceeding to the next refinement level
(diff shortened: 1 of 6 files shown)


src/Main/EvolveLevel.cpp
```diff
@@ -749,6 +767,21 @@ void EvolveLevel( const int lv, const double dTime_FaLv )
       } // if ( lv != TOP_LEVEL  &&  AdvanceCounter[lv] % REGRID_COUNT == 0 )
 // ===============================================================================================
 
+
+//    14. user-specified operations before proceeding to the next refinement level
```
hfhsieh, line 771: before proceeding to the next `substep`?

hfhsieh (commented):
Look good. One minor question on the description.

PR #145, opened 2022-09-12
Update Taiwania3 job scripts

JeiwHuang:
1. Exclude the large-memory nodes.
Memory size: cpn[3001-3120] are 768GB and cpn[3241-3360] are 384GB

2. Set the memory limit per node according to this [Q&A](https://iservice.nchc.org.tw/nchc_service/nchc_service_qa_single.php?qa_code=689).


hyschive (approved):
@JeiwHuang Thanks for taking care of it!

PR #152, opened 2022-11-02
Pass `Emag` to `Flu_ResetByUser_Func_Ptr()`

hyschive:


Inhyeok-Song:
You might need to change `Flu_ResetByUser_Func_Ptr` in `src/Model_Hydro/Hydro_Init_ByFunction_AssignData.cpp`

I have one question about this.
`Hydro_Init_ByFunction_AssignData` always invokes `Flu_ResetByUser_Func_Ptr` with `AuxArray=NULL_REAL`. Thus if we use `AuxArray`, `Flu_ResetByUser_Func_Ptr` in `Hydro_Init_ByFunction_AssignData`  would access NULL_REAL and gives `segmentation fault`.
Since we haven't use any `AuxArray` in `Flu_ResetByUser_Func_Ptr` this didn't happen.
So I wonder if this is expected or a bug

hyschive:
@Inhyeok-Song Good point! I've fixed it.

src/Model_Hydro/Hydro_Init_ByFunction_AssignData.cpp
```diff
@@ -248,8 +248,9 @@ void Hydro_Init_ByFunction_AssignData( const int lv )
             Init_Function_User_Ptr( fluid_sub, x, y, z, Time[lv], lv, NULL );
 
 //          modify the initial condition if required
+//          --> always set the magnetic energy to zero since fluid_sub[ENGY] doesn't include that
             if ( OPT__RESET_FLUID )
-               Flu_ResetByUser_Func_Ptr( fluid_sub, x, y, z, Time[lv], 0.0, lv, NULL );
+               Flu_ResetByUser_Func_Ptr( fluid_sub, (real)0.0, x, y, z, Time[lv], 0.0, lv, NULL );
```
Inhyeok-Song, line 253: It's good to set `Emag=0.0`.

But what I meant is that this function here always takes `AuxArray[] = NULL;`
Flu_ResetByUser_Func_Ptr( fluid_sub, (real)0.0, x, y, z, Time[lv], 0.0, lv, `NULL` );

So, I am worrying that,  in the future, if we use `Flu_ResetByUser_Func_Ptr` somewhere else with `AuxArray != NULL`.
`Flu_ResetByUser_Func_Ptr` will access `NULL` when it is calling from `Hydro_Init_ByFunction_AssignData`
and will give us a `segmentation fault`.

hyschive: @Inhyeok-Song `AuxArray[]` is user-defined. To use that, users must define their own `Flu_ResetByUser_API_Ptr` instead of using the default one. Note that `Flu_ResetByUser_API_Default()` also has `AuxArray = NULL`.

hyschive: @Inhyeok-Song My comment above is not fully correct. Even with a user-defined `Flu_ResetByUser_API_Ptr()`, `Hydro_Init_ByFunction_AssignData()` will still set `AuxArray=NULL` as you mentioned. So I think to use `AuxArray` one must also overwrite the default `Hydro_Init_ByFunction_AssignData()`, which is however not supported it. I'll file an issue about that. Thanks for the careful check!

hyschive: @Inhyeok-Song #153

hyschive:
@Inhyeok-Song Thanks for the check again. But I don't think we need to do that. We will either remove or update those deprecated tests in the future. They are currently not compiled at all so shouldn't cause any problems.

src/Fluid/Flu_ResetByUser.cpp
```diff
@@ -132,19 +136,19 @@ void Flu_ResetByUser_API_Default( const int lv, const int FluSg, const double Ti
 
          for (int v=0; v<NCOMP_TOTAL; v++)   fluid[v] = amr->patch[FluSg][lv][PID]->fluid[v][k][j][i];
 
+#        ifdef MHD
```
hfhsieh, line 139: Just notice that we implicitly assume users will enable `MHD` only in `MODEL = HYDRO`, and ignore the check `MODEL = HYDRO` as the code here. However, I do not find any check in `Aux_Check_Parameter.cpp` for that. Should we add it?

hfhsieh (approved):
Look goods. Just one unrelated comment.

src/Fluid/Flu_ResetByUser.cpp
hyschive: @hfhsieh I don't think it's necessary since one can only enable `MHD` when adopting `MODEL=HYDRO` in `Makefile` ([here](https://github.com/gamer-project/gamer/blob/master/src/Makefile#L53)).

Inhyeok-Song (approved):

PR #154, opened 2022-11-05
Recursive refine

hyschive:
Add the runtime option [REFINE_NLEVEL](https://github.com/gamer-project/gamer/wiki/Runtime-Parameters%3A-Refinement#REFINE_NLEVEL) to allow for refining multiple levels at once during refinement.


KunkelAlexander (approved):
Everything looks fine: I checked that the flag appears in the record files, is loaded correctly from the HDF5-file and that multiple refinements at once can be triggered.

hyschive:
@KunkelAlexander Thanks for the review! I'll merge it right away.

PR #167, opened 2023-01-20
Find extrema

hyschive:
### Summary

Add `Aux_FindExtrema()` to find the value and location of the extreme value of a given field within a target spherical region.

### Features and limitations:
- Support intrinsic fluid fields and gravitational potential
- Does not support magnetic field, derived fields, and deposited particle fields
- Does not support computing multiple fields at once
- Support periodic BC
- Support finding either the maximum or minimum value
- Support hybrid OpenMP/MPI parallelization

### Future works
- Support more fields
- Support computing center-of-mass

@hfhsieh Please help review it when you have time. Thanks!
Once merged, we can apply it in `Record_CCSN_CentralQuant()`.


include/Extrema.h
```diff
@@ -0,0 +1,132 @@
...
+//                CreateMPIType : Create a MPI derived datatype for struct Extrema_t
+//-------------------------------------------------------------------------------------------------------
+struct Extrema_t
+{
+
+// data members
+// --> must be consistent with CreateMPIType()
+// ===================================================================================
+// input parameters
+   int    Field;
```
hfhsieh, line 33: The datatype of `Field` should be `long`.

include/Extrema.h
```diff
@@ -0,0 +1,132 @@
...
+   // Description :  Constructor of the structure "Extrema_t"
+   //
+   // Note        :  Initialize the data members
+   //
+   // Parameter   :  None
+   //===================================================================================
+   Extrema_t()
+   {
+
+      Field     = -1;
```
hfhsieh, line 57: The initial value of `Field` should be 0 (`= _NONE`), otherwise the check [here](https://github.com/hyschive/gamer-fork/blob/pr-find-extrema/src/Auxiliary/Aux_FindExtrema.cpp#L82) has no effect.

include/Extrema.h
```diff
@@ -0,0 +1,132 @@
...
+   // Parameter   :  MPI_Extrema_t : MPI derived datatype to be returned
+   //
+   // Return      :  MPI_Extrema_t
+   //===================================================================================
+   void CreateMPIType( MPI_Datatype *MPI_Extrema_t )
+   {
+
+      const int          NBlk         = 9;
+      const int          Length[NBlk] = { 1, 1, 3, 1, 3, 1, 1, 1, 3 };
+      const MPI_Datatype Type  [NBlk] = { MPI_INT, MPI_DOUBLE, MPI_DOUBLE, MPI_GAMER_REAL, MPI_DOUBLE, MPI_INT, MPI_INT, MPI_INT, MPI_INT };
```
hfhsieh, line 107: Modify the first element to `MPI_LONG` if the datatype of `Field` is updated to `long`.

src/Auxiliary/Aux_FindExtrema.cpp
```diff
@@ -0,0 +1,315 @@
...
+//          loop over all cells
+            const real (*FluidPtr)[PS1][PS1][PS1] = amr->patch[ amr->FluSg[lv] ][lv][PID]->fluid;
+#           ifdef GRAVITY
+            const real (*PotPtr  )[PS1][PS1]      = amr->patch[ amr->PotSg[lv] ][lv][PID]->pot;
+#           endif
+            const double x0  = amr->patch[0][lv][PID]->EdgeL[0] + 0.5*dh;
+            const double y0  = amr->patch[0][lv][PID]->EdgeL[1] + 0.5*dh;
+            const double z0  = amr->patch[0][lv][PID]->EdgeL[2] + 0.5*dh;
+
+            for (int k=0; k<PS1; k++)  {  const double z = z0  + k*dh;
```
hfhsieh, line 211: Remove one redundant space before `+ k*dh`

src/Auxiliary/Aux_FindExtrema.cpp
```diff
@@ -0,0 +1,315 @@
...
+                     OMP_Extrema[TID].Level    = lv;
+                     OMP_Extrema[TID].PID      = PID;
+                     OMP_Extrema[TID].Cell[0]  = i;
+                     OMP_Extrema[TID].Cell[1]  = j;
+                     OMP_Extrema[TID].Cell[2]  = k;
+                  }
+               } // if ( r2 < MaxR2 )
+            }}} // i,j,k
+         } // for (int PID=0; PID<amr->NPatchComma[lv][1]; PID++)
+      } // for (int lv=lv_min; lv<=lv_max; lv++)
```
hfhsieh, line 262: Should be `for (int lv=MinLv; lv<=MaxLv; lv++)`. Same in `Aux_ComputeProfile()`

src/Auxiliary/Aux_FindExtrema.cpp
```diff
@@ -0,0 +1,315 @@
...
+   Extrema->Field     = Field;
+   Extrema->Radius    = MaxR;
+   Extrema->Center[0] = Center[0];
+   Extrema->Center[1] = Center[1];
+   Extrema->Center[2] = Center[2];
+
+
+// for MPI only
+#  ifndef SERIAL
+// define an MPI dervied datatype for Extrema_t
```
hfhsieh, line 286: typos: `dervied` -> `derived`

hfhsieh (changes requested):
@hyschive Looks good. This version is more sophiscated than the one I implemented in the PR for leakage scheme. Some additional comments are:

- It is better to use `>= / <=`, instead of `> / <`, to avoid inconsistent results among MPI processes when there are multiple cells have the same extrema.
- It could be better to have a bool arguement to control whether the coordinate is shifted to the box center if `x, y, z = 0.5 * dh`. Like the code [here](https://github.com/hfhsieh/gamer-fork/blob/ccsn_leakage_tentative/src/Auxiliary/Aux_GetCoord_Field.cpp#L133-L140)
- I am not sure whether the creation and destruction of MPI datatype are time-consuming. If it is, maybe we can do that once before and after the main routine.

hyschive:
@hfhsieh Thanks for the very careful review! I've incorporated all inline comments.

> * It is better to use `>= / <=`, instead of `> / <`, to avoid inconsistent results among MPI processes when there are multiple cells have the same extrema.

Are you referring to these [lines](https://github.com/hyschive/gamer-fork/blob/pr-find-extrema/src/Auxiliary/Aux_FindExtrema.cpp#L301-L302)? Note that unlike `Record_CCSN_CentralQuant()`, here only the root rank will do the comparison and then broadcast results to all ranks. So the results should be identical among all ranks.

> * It could be better to have a bool arguement to control whether the coordinate is shifted to the box center if `x, y, z = 0.5 * dh`. Like the code [here](https://github.com/hfhsieh/gamer-fork/blob/ccsn_leakage_tentative/src/Auxiliary/Aux_GetCoord_Field.cpp#L133-L140)

Since this function is not for a specific test problem, I would prefer to keep the coordinates at the cell center to respect the original simulation data. Users can shift the returned coordinates to the box center manually if preferred.

> * I am not sure whether the creation and destruction of MPI datatype are time-consuming. If it is, maybe we can do that once before and after the main routine.

I don't think it could be that time-consuming as long as there are enough cells. So probably no need to worry about that.

include/Extrema.h
```diff
@@ -54,22 +54,22 @@ struct Extrema_t
    Extrema_t()
    {
 
-      Field     = -1;
+      Field     = _NONE;
```
hfhsieh, line 57: Since `Macro.h` is not included in this file, I think `_NONE` is not defined yet.

hyschive: @hfhsieh Fixed

hfhsieh (approved):

PR #179, opened 2023-02-06
Update New libyt API and Add Interactive Mode

cindytsai:
# Update New libyt API and Add Interactive Mode

## Update to New libyt API
See https://github.com/calab-ntu/libyt/pull/70

## Add Interactive Mode and Reloading Script Feature
- Prerequisite:
  - Must compile `libyt` with `-DINTERACTIVE_MODE`.
  - Set `-DSUPPORT_LIBYT_INTERACTIVE` in GAMER `Makefile`.
- Feature:
  - Does not shutdown the simulation when there are errors in in situ analysis using python.
  - Access python prompt and get feedback in in situ analysis. Changes will maintain throughout the following steps.


hyschive:
@cindytsai Please inform us when this PR is ready for review.

@KunkelAlexander Could you also help review it? Thanks!

cindytsai:
@hyschive @KunkelAlexander This is ready to review, thanks.

KunkelAlexander:
@cindytsai I am very sorry to bother you, but I still could not get the latest version of `libyt` from `calab-ntu` to work with this pull request (using Python 3.7). I am trying to run the `Plummer` test in default configuration, but get the error `Importing YT inline analysis script "inline_script" ... failed (please do not include the ".py" extension)!`. The have not modified the default Plummer test, so I assume it should work out of the box. Do you have an idea why it may not work?

As for the code modifications in this pull request, everything looks fine. I just need to test it. 

As for the installation of `libyt`, I found that I head to install the `readline` library as an external dependency when turning the `-DINTERACTIVE_MODE` on. You only use it in one place, so maybe you can do without it. Otherwise, you should mention it explicitly as a dependency that does not come with `build-essentials`, for instance. Furthermore, I had to add `-std=c++11` to`libyt`'s `CXXFLAGs`. Maybe, you can add this line to the default makefile in case the compiler requires it for C++11 support.

cindytsai:
> EDIT: I could get the code to work now by including the shared library provided by gcc (gamer finally threw the error version CXXABI_1.3.9 not found in debug mode which got me on the right track). :).

:+1: glad you fixed it.

> Otherwise, you should mention `readline` explicitly as a dependency that does not come with `build-essentials`.

I thought `readline` is shipped as a standard library on most Linux and FreeBSD system. ([link](https://tiswww.case.edu/php/chet/readline/rltop.html#Distributions))
 - Could you tell me what operating system are you using?

> Maybe one comment regarding the YT_INFO and YT_DEBUG messages.

For `YT_DEBUG`, because different MPI process might give different arguments, I prefer to stay as it is.

For `YT_INFO`, I'm still thinking about should I change its overall settings by printing on root only, or specifically mark it to print on root at some points. I'm still thinking on this (that is for later updates, cause the latest `PR` is to pin down API names), but thanks for pointing out.

KunkelAlexander:
@cindytsai 
Another suggestion: I think you should check for the minimum required `yt` version for the latest `libyt` version to work. 
I tested the `example` shipped with `libyt` using my own and Guan-Ming's previous `libyt` installations and everything works as expected ( Python 3.7, SHA of libyt `011fa07`). But with the latest version of `libyt `, I get the following error:

>     self.grid_particle_count[:] = hierarchy['grid_particle_count'][:]
> KeyError: 'grid_particle_count'
> [YT_ERROR  ] Invoking "yt_inline_ProjectionPlot('density')" ... failed
>              ==> file <yt_run.cpp>, line <91>, function <yt_run_FunctionArguments>

In your commit history to your yt - fork , I see that yt removed `grid_particle_count`. I took me some time to figure out that we should have pulled the latest version of your yt-fork as well. 

As for `readline`, it seems like it is not shipped with Ubuntu (and therefore Debian I guess) (https://askubuntu.com/questions/194523/how-do-i-install-gnu-readline). 

As for C++11-support, g++ on Ubuntu 20.04 required the flag (I do not have my private laptop here right now, so I cannot tell you the exact version). I think a pull request is a little too complicated for one line of code ;). You can simply add it if you like and if it does not cause problems.

cindytsai:
@KunkelAlexander 
I think I have mentioned pulling my `yt` repo `libyt` branch at Readme in `libyt`, but I should have mentioned here again. Sorry about that,. I will PR to main `yt-project/main` branch recently. This is just a temporary alternative. 

> it seems like it is not shipped with Ubuntu (and therefore Debian I guess)
I'll mention that in the next update. Thanks!
Just curious, do you install `readline` when you try to run `libyt`? (I have tested on Ubuntu 20.04, 22.04 and CentOS 7.9 (eureka), all of them shipped with `readline`. Not sure about Debian though.)

KunkelAlexander:
@cindytsai 
Yes, you have mentioned pulling your `yt` repo. But what I mean is that even then, there is a minimal version of your `yt` repo required to work, right? Anyway, I should have checked earlier. My bad!
As for `readline`, maybe it was installed through other packages? I installed a fresh version of ubuntu 22.04 last week and it was also not preinstalled if I am not mistaken. After installing the appropriate package, it worked.

hyschive:
@cindytsai If there is a minimal version of your `yt` repo required to work, as @KunkelAlexander mentioned, it will be better to validate the `yt` version when initializing `libyt` and raise an error message when it is violated. Most people won't read README carefully.

cindytsai:
@KunkelAlexander Do you mean Python's minimal version, or yt's? Anyhow, `libyt` probably be singled out, so all `yt` version that uses Python 3.7 to 3.10 (3.6 may also worked but not tested in GitHub Action) can use `libyt`.

KunkelAlexander:
@cindytsai 
I am specifically referring to your commit 
`[Fix bug, libyt.hierarchy does not have grid_particle_count now.](https://github.com/cindytsai/yt/commit/58343d94a2c65e8a4d1175b197728219f87938f2)  `. 
With an old version of libyt and an old version of your libyt branch of the yt library, everything worked fine. However, updating libyt broke the configuration since the interfaces changed. As a consequence, I also had to pull the latest version of the libyt branch of the yt library. I was just wondering if a version check could help eliminate such a problem.  You could check something like 
`if libyt version > X,` you need at least commit Y of the libyt branch of yt (or at least yt version Z once your version of yt is merged into the main branch). 

Oh, another thing. I finally got `example` to run with the latest `libyt` version. Have you mentioned anywhere that `mpi4py` is also a required dependency to run this example?

KunkelAlexander (approved):
@cindytsai @hyschive 
The pull request works nicely and I can access the `libyt` interactive shell during a `GAMER` run. The code modifications themselves also look good.

KunkelAlexander:
@cindytsai 
One more question regarding the interactive mode: 
You write that it `Does not shutdown the simulation when there are errors in in situ analysis using python.` 
But when I actually added a new function with an error to the analysis pipeline through the interactive shell, the program crashed. Is that expected behaviour? 

> 
> [YT_INFO   ] Performing YT inline analysis func(a ,2 ,"3" ) ...
> [YT_ERROR  ] SyntaxError: invalid syntax (<string>, line 2)
> [YT_ERROR  ] Error occurred on rank 0.
> [YT_ERROR  ] Invoking "func(a ,2 ,"3" )" ... failed
>              ==> file <func_status_list.cpp>, line <174>, function <run_func>
> [YT_ERROR  ] Something went wrong when running new added functions
>              ==> file <yt_run_InteractiveMode.cpp>, line <33>, function <yt_run_InteractiveMode>
> ********************************************************************************
> ERROR : yt_run_InteractiveMode() failed !!
>         Rank <0>, file <YT/YT_Inline.cpp>, line <255>, function <YT_Inline>
> ********************************************************************************
> 
> Program terminated with error ...... rank 0

cindytsai:
@KunkelAlexander 
Thank you! This is a new emerging bug while I was fixing something else. The stuff you enter (I supposed it's from libyt doc) has been tested before, but later, I changed the implementation. (Too bad, wrong decision at that time. 😓 )

Anyway, this is caused by `"`, I will fix that tomorrow. It should work if you change `"` to `'`. This has nothing to do with GAMER. I'll fix this alongside with supporting C library, then PR to libyt. (Not sure if I can finish them this week, and not sure if you want to wait for it.)

hyschive:
@cindytsai Have you fixed all the issues raised by @KunkelAlexander?  If so, can you update this PR to match the latest `main` branch before I review it?

cindytsai:
@hyschive Yes, I have fixed it [here](https://github.com/calab-ntu/libyt/issues/72). 
I have merged newest main branch to this PR and tested it. It looks fine.

src/Makefile
```diff
@@ -219,10 +219,15 @@ SIMU_OPTION += -DOPENMP
...
 
-# use patch group as the unit in libyt; this will speed up inline-analysis but increase memory consumption
+# use patch group as the unit in libyt, this will speed up inline-analysis but increase memory consumption
 # --> must enable SUPPORT_LIBYT
 #SIMU_OPTION += -DLIBYT_USE_PATCH_GROUP
 
+# support libyt interactive mode, this activate python prompt and does not shut down simulation when there are
+# errors in inline python script. must compile libyt with -DINTERACTIVE_MODE
+# --> must enable SUPPORT_LIBYT
+#SIMU_OPTION += -DSUPPORT_LIBYT_INTERACTIVE
```
hyschive, line 229: I would suggest simplifying the option name as `LIBYT_INTERACTIVE`, which also conforms with the naming convention `SUPPORT_SOFTWARE_NAME` adopted here.

include/Global.h
```diff
@@ -7,6 +7,9 @@
 #include "AMR.h"
 #ifdef SUPPORT_LIBYT
 #include <libyt.h>
+#ifdef SUPPORT_LIBYT_INTERACTIVE
+#include <libyt_interactive_mode.h>
+#endif
```
hyschive, line 12: Move it to `include/GAMER.h` ([here](https://github.com/gamer-project/gamer/blob/main/include/GAMER.h#L57)). Also, check my general comment about including `libyt_interactive_mode.h` inside `lbiyt.h`.

src/YT/YT_Inline.cpp
```diff
@@ -237,19 +236,27 @@ void YT_Inline()
...
-   if ( yt_inline( "yt_inline" ) != YT_SUCCESS )   Aux_Error( ERROR_INFO, "yt_inline() failed !!\n" );
+// 5-1. perform yt inline analysis
+   if ( yt_run_FunctionArguments( "yt_inline_inputArg", 1, "\'Dens\'" ) != YT_SUCCESS )
+       Aux_Error( ERROR_INFO, "yt_run_FunctionArguments() failed !!\n" );
+   if ( yt_run_Function( "yt_inline" ) != YT_SUCCESS )
+       Aux_Error( ERROR_INFO, "yt_run_Function() failed !!\n" );
+
+#ifdef SUPPORT_LIBYT_INTERACTIVE
+// 5-2. activate libyt interactive mode
+   if ( yt_run_InteractiveMode("LIBYT_STOP") != YT_SUCCESS )
```
hyschive, line 254: Do you prefer letting each simulation code define its own file name (i.e., `LIBYT_STOP` here) or fixing it to be `LIBYT_STOP` by `libyt`?

hyschive (commented):
@cindytsai Very impressive! I enjoy the new Python prompt and how users can modify existing or add new inline functions!

General comments:
- [ ] Why not include `libyt_interactive_mode.h` in `libyt.h` so that simulation codes only need to include `libyt.h` even when enabling `INTERACTIVE_MODE`?
- [ ] When running batch instead of interactive jobs (i.e., `qsub job_script` instead of `qsub -I`), is it still possible to reload a yt script based on this PR? IIRC, `LIBYT_STOP` only works for interactive jobs. If it is not possible for now, consider implementing this feature in the future since production scientific simulations use batch instead of interactive jobs.

Minor comments not directly related to this PR:
- [ ] Installation of `libyt`
  - [ ] How about doing `cp libyt.so* ../lib/` automatically at the end of `Make`?
  - [ ] Clarify how to determine `PYTHON_PATH, PYTHON_VERSION, NUMPY_PATH`.
  - [ ] Do you think it would be helpful to validate the minimum required versions of `yt` and `python` as @KunkelAlexander suggests?
- [ ] User guide 
  - [ ] Remove `/` from the beginning of, for example, `/libyt/src`, `/libyt/lib`, and `/libyt/include` in the installation guide.
  - [ ] Add `cd yt` before `pip install .` [here](https://github.com/calab-ntu/libyt#yt).
  - [ ] For `%libyt load` [here](https://github.com/calab-ntu/libyt/blob/main/doc/InteractivePythonPrompt.md#load), change to "All NEW functionS detected ..."
  - [ ] Have you mentioned the usage of `LIBYT_STOP` in the `libyt` user guide?
  - [ ] Provide some concrete examples of the usage and results of different `%libyt [Command]` for [Interactive Python Prompt](https://github.com/calab-ntu/libyt/blob/main/doc/InteractivePythonPrompt.md#interactive-python-prompt)
  - [ ] I don't quite understand the following comment [here](https://github.com/calab-ntu/libyt/blob/main/doc/InteractivePythonPrompt.md#run) (and fix the typo `was` --> `were`).
> When calling [yt_run_InteractiveMode](https://github.com/calab-ntu/libyt/blob/main/doc/ActivateInteractiveMode.md#yt_run_interactivemode), libyt will run all the functions that was set to run using [%libyt run](https://github.com/calab-ntu/libyt/blob/main/doc/InteractivePythonPrompt.md#run-1), but had not been run by [yt_run_Function](https://github.com/calab-ntu/libyt/blob/main/doc/PerformInlineAnalysis.md#yt_run_function) or [yt_run_FunctionArguments](https://github.com/calab-ntu/libyt/blob/main/doc/PerformInlineAnalysis.md#yt_run_functionarguments) yet.

hyschive:
> When running batch instead of interactive jobs (i.e., qsub job_script instead of qsub -I), is it still possible to reload a yt script based on this PR? IIRC, LIBYT_STOP only works for interactive jobs. If it is not possible for now, consider implementing this feature in the future since production scientific simulations use batch instead of interactive jobs.

@cindytsai Regarding this, how about adding another file `LIBYT_RELOAD`, putting all the Python statements therein, and executing it directly without activating a Python prompt? For example, we can put the following codes in `LIBYT_RELOAD`.

```
%libyt load filename
%libyt idle func1
a = 1
def func2(*args):
%libyt run func2 a 2 "3"
```

src/YT/YT_Inline.cpp
cindytsai: Yes

hyschive:
@cindytsai PR #193 can be used together with `LIBYT_RELOAD` to _reload_ a `libyt` script in the following procedure.
1. A user realizes something in the original `libyt` script needs to be revised.
2. Create the file `PAUSE_GAMER_PAUSE` to pause a run.
3. Revise the `libyt` script.
4. Create the file `LIBYT_RELOAD`.
5. Delete `PAUSE_GAMER_PAUSE` to resume a simulation, which will then reload the revised `libyt` script when detecting `LIBYT_RELOAD`.

What do you think?

include/Global.h
cindytsai: I will consider this later, currently we (Matt and I) are working on Enzo, and there are other minor things to do.
Do you want to put it as an individual issue in libyt?

hyschive:
@cindytsai I've moved things not directly related to this PR to the following three issues so that they won't block this PR. I'll merge it once all the inline comments are resolved.
- calab-ntu/libyt#75
- calab-ntu/libyt#76
- gamer-project/gamer#194

src/Makefile
cindytsai: done

cindytsai:
> @cindytsai PR #193 can be used together with `LIBYT_RELOAD` to _reload_ a `libyt` script in the following procedure.
> 
> 1. A user realizes something in the original `libyt` script needs to be revised.
> 2. Create the file `PAUSE_GAMER_PAUSE` to pause a run.
> 3. Revise the `libyt` script.
> 4. Create the file `LIBYT_RELOAD`.
> 5. Delete `PAUSE_GAMER_PAUSE` to resume a simulation, which will then reload the revised `libyt` script when detecting `LIBYT_RELOAD`.
> 
> What do you think?

(@hyschive Sorry, I'm really busy lately.)
I think implementing SSH would be better. I also has plans for Jupyter Notebook in the future.
But yes, this is feasible, but I think this should be a single API.

cindytsai:
@matthewturk Yeah, I think I can follow [this](https://api.libssh.org/master/libssh_tutor_guided_tour.html)

cindytsai:
@matthewturk 
1. Nope 😓 
2. OK! That SSH was just for that ugly raw prompt. (Beside the point) For jupyter notebook, I think (plan) I would use [Xeus](https://xeus.readthedocs.io/en/latest/). Want to make sure that it is feasible and useful. Cause it can directly communicate to C and Python. The major communication is still using MPI on C side, and whatever it is at Python side (can be dask etc).

hyschive (approved):

PR #184, opened 2023-03-03
Fix memory leak and check cubic box in `Output_BasePowerSpectrum.cpp`

koarakawaii:
* Fix memory leak in `Output_BasePowerSpectrum.cpp`
* Add comment in `Patch2Slab` function
* Add cubic box check in `Output_BasePowerSpectrum.cpp`


hyschive (approved):
@koarakawaii @JeiwHuang Thanks for fixing this memory leakage problem!

PR #210, opened 2023-06-14
Initialize source-term fields

hyschive:
Call `Src_AdvanceDt()` at the end of `Init_GAMER()` to initialize all source-term fields (e.g., cooling time).

### DIFF
--- a/src/Init/Init_GAMER.cpp
+++ b/src/Init/Init_GAMER.cpp
@@ -320,17 +320,17 @@ void Init_GAMER( int *argc, char ***argv )
 
 // initialize source-term fields (e.g., cooling time)
 // --> necessary for, for example, estimating the source-term time-step at the first step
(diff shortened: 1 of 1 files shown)


src/Init/Init_GAMER.cpp
```diff
@@ -318,4 +318,33 @@ void Init_GAMER( int *argc, char ***argv )
 #  endif // #ifdef PARTICLE
 
 
+// initialize source-term fields (e.g., cooling time)
+// --> necessary for, for example, estimating the source-term time-step at the first step
+// --> must ensure each source term does not modify any fluid field (e.g., gas internal energy) when dt=0.0
```
hfhsieh, line 323: Remove the trailing spaces at lines 323, 326, and 333

hfhsieh (approved):
Looks good to me.

PR #223, opened 2023-07-08
Add rpath in Makefile_base

hyschive:
Add `rpath` in `Makefile_base` for all library paths so that users do not have to set `LD_LIBRARY_PATH` manually.

Note that it doesn't mean `rpath` has higher precedence than `LD_LIBRARY_PATH` since whether `rpath` sets `RPATH` or `RUNPATH` is system-dependent. See the following references for details. As a result, if users still set `LD_LIBRARY_PATH`, they must make sure that it is consistent with the paths in `MACHINE.config`. 

References:
- [OpenMPI manual](https://docs.open-mpi.org/en/v5.0.x/installing-open-mpi/configure-cli-options/installation.html#installation-options)
- [wiki](https://en.wikipedia.org/wiki/Rpath)

### DIFF
--- a/src/Makefile_base
+++ b/src/Makefile_base
@@ -369,7 +369,7 @@ endif
 # -------------------------------------------------------------------------------
 ifeq "$(filter -DGPU, $(SIMU_OPTION))" "-DGPU"
 LIB += -L$(CUDA_PATH)/lib64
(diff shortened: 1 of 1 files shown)


hfhsieh (approved):
Looks good to me.

PR #225, opened 2023-07-13
Reset B fields

hyschive:
### New features
- Support resetting magnetic fields during the runtime and initialization
- Support assigning magnetic fields directly or using a vector potential
- Support resetting magnetic fields in the test problem `BlastWave`
- See [wiki](https://github.com/gamer-project/gamer/wiki/Adding-New-Simulations#fields-resetting) for instructions

### Demos
- Resetting magnetic fields during `1e-3 <= t <= 3e-3` within a radius of roughly `1e-1` in the blast-wave test:
https://github.com/gamer-project/gamer/assets/18057331/bf7df82c-c174-4d7e-9a7f-875fb561c6a1

- Relative divergence-free errors in three different scenarios for the blast-wave test:
  - Green: Using magnetic fields directly without touching any AMR resolution interfaces --> large errors
  - Blue: Using a vector potential without touching any AMR resolution interfaces --> machine precision
  - Red: Using a vector potential that touches AMR resolution interfaces --> moderate errors
fig__divergence_free_errors


src/Model_Hydro/Hydro_Init_ByFunction_AssignData.cpp
```diff
@@ -186,31 +208,76 @@ void Hydro_Init_ByFunction_AssignData( const int lv )
...
+            if ( i != PS1 )   AxTID[idx] = MHD_ResetByUser_VecPot_Ptr( x+dh_2, y,      z,      0.0, 0.0, lv, 'x', NULL );
+            if ( j != PS1 )   AyTID[idx] = MHD_ResetByUser_VecPot_Ptr( x,      y+dh_2, z,      0.0, 0.0, lv, 'y', NULL );
+            if ( k != PS1 )   AzTID[idx] = MHD_ResetByUser_VecPot_Ptr( x,      y,      z+dh_2, 0.0, 0.0, lv, 'z', NULL );
+
+            idx ++;
+         }}} // i,j,k
+      } // if ( UseVecPot )
+
+
       if ( !OPT__INIT_BFIELD_BYVECPOT ) {
```
hfhsieh, line 264: - The code for resetting the initial B field is within this `if` block, which will be executed only if the B field is not initialized from vector potential (`OPT__INIT_BFIELD_BYVECPOT = 0`). Maybe we need to mention that in the wiki, update the condition as `if ( !OPT__INIT_BFIELD_BYVECPOT  ||  ResetMag )`, or move the inner `if ( ResetMag )` block out.
- The code in lines 239-261 for initializing `A[x/y/z]TID` is used only when `ResetMag = True`. Maybe we can move it into the `if ( ResetMag )` block.

src/TestProblem/Hydro/BlastWave/Init_TestProb_Hydro_BlastWave.cpp
```diff
@@ -152,6 +173,14 @@ void SetParameter()
                                                                                        Blast_Center[2] );
 #     ifdef MHD
       Aux_Message( stdout, "  magnetic field strength   = %13.7e\n", Blast_BField );
+      Aux_Message( stdout, "  resetting magnetic field  = %d\n",     OPT__RESET_FLUID );
+      if ( OPT__RESET_FLUID ) {
```
hfhsieh, line 177: Not sure if we also need to include `OPT__RESET_FLUID` in `if (...)`.

hfhsieh (approved):
Looks good to me. I have left some comments.

src/TestProblem/Hydro/BlastWave/Init_TestProb_Hydro_BlastWave.cpp
hfhsieh: Should be `OPT__RESET_FLUID_INIT`

src/Model_Hydro/MHD_ResetByUser.cpp
```diff
@@ -0,0 +1,214 @@
...
+   if ( UseVecPot )
+   {
+      switch ( Component )
+      {
+         case 'x' :  B_out = B_in + dt*MHD_ResetByUser_A2Bx( Ax, Ay, Az, i, j, k, dh );    break;
+         case 'y' :  B_out = B_in + dt*MHD_ResetByUser_A2By( Ax, Ay, Az, i, j, k, dh );    break;
+         case 'z' :  B_out = B_in + dt*MHD_ResetByUser_A2Bz( Ax, Ay, Az, i, j, k, dh );    break;
+         default  :  Aux_Error( ERROR_INFO, "unsupported component (%c) !!\n", Component );
+      }
+   } // if ( UseVecPot )
```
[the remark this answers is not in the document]
hyschive: I prefer to keep it for clarity.

src/TestProblem/Hydro/BlastWave/Init_TestProb_Hydro_BlastWave.cpp
hyschive: This test is only for testing `OPT__RESET_FLUID` but not `OPT__RESET_FLUID_INIT`. The latter has no effect here since `dt==0.0`.

src/Model_Hydro/Hydro_Init_ByFunction_AssignData.cpp
hyschive: @hfhsieh 
- Good catch! I've fixed it. Now we can reset B fields even with `OPT__INIT_BFIELD_BYVECPOT`.
- It should be fine since `UseVecPot` is true only when `ResetMag` is true. Note that for clarity I've renamed `UseVecPot` as `ResetMag_UseVecPot`.

PR #226, opened 2023-07-17
Update FFTW and GSL interface

KunkelAlexander:
This pull request updates the FFTW interface and introduces a new GSL interface in order to make it easier to switch between single and double precision FFTW and GSL routines.


hyschive (approved):
@KunkelAlexander Looks good :)

PR #229, opened 2023-08-18
Support large FFT

hyschive:
### Modifications
- Replace `int` with `long` in FFT to utilize the 64-bit transformation in FFTW3 ([ref](https://www.fftw.org/fftw3_doc/64_002dbit-Guru-Interface.html))
- Check integer overflow for large FFT
- Convert `local_nz` to integer for `MPI_Allgather()`

### Tests
- `Plummer` with `NX0_TOT_X/Y/Z = 1280` and `MAX_LEVEL=0` using 6 MPI processes
- Original code fails due to integer overflow for both FFTW2 and 3. Increasing the number of MPI processes to 12 solves the problem.
- Modified code works well for FFTW3 and raises the integer overflow error message for FFTW2.

### DIFF
--- a/src/Output/Output_BasePowerSpectrum.cpp
+++ b/src/Output/Output_BasePowerSpectrum.cpp
@@ -58,10 +58,17 @@ void Output_BasePowerSpectrum( const char *FileName, const long TVar )
 #  endif // #  if ( SUPPORT_FFTW == FFTW3 ) ... # else
 #  endif // #  ifdef SERIAL ... # else
 
-// check integer overflow (in a non-rigorous way)
-   if ( total_local_size <= 0 )
-      Aux_Error( ERROR_INFO, "total_local_size (%d) <= 0 (local_nx/y/z = %d, %d, %d) for FFT !!\n",
-                 total_local_size, local_nx, local_ny, local_nz );
+// check integer overflow (assuming local_nx*local_ny*local_nz ~ total_local_size)
+   const long local_nxyz = (long)local_nx*(long)local_ny*(long)local_nz;
+
+   if ( local_nx < 0 || local_ny < 0 || local_nz < 0 )
+      Aux_Error( ERROR_INFO, "local_nx/y/z (%ld, %ld, %ld) < 0 for FFT !!", local_nx, local_ny, local_nz );
+
+   if (  ( sizeof(mpi_index_int) == sizeof(int) && local_nxyz > __INT_MAX__ )  ||  total_local_size < 0  )
+      Aux_Error( ERROR_INFO, "local_nx*local_ny*local_nz = %d*%d*%d = %ld > __INT_MAX__ (%d)\n"
+                     "        and/or total_local_size (%ld) < 0 for FFT, suggesting integer overflow !!\n"
+                     "        --> Try using more MPI processes\n",
+                 local_nx, local_ny, local_nz, local_nxyz, __INT_MAX__, total_local_size );
 
 // collect "local_nz" from all ranks and set the corresponding list "List_z_start"
    int List_nz     [MPI_NRank  ];   // slab thickness of each rank in the FFTW slab decomposition
--- a/src/Output/Output_BasePowerSpectrum.cpp
+++ b/src/Output/Output_BasePowerSpectrum.cpp
@@ -81,11 +81,11 @@ void Output_BasePowerSpectrum( const char *FileName, const long TVar )
    const int NRecvSlice = MIN( List_z_start[MPI_Rank]+local_nz, NX0_TOT[2] ) - MIN( List_z_start[MPI_Rank], NX0_TOT[2] );
 
    double *PS_total     = NULL;
-   real   *VarK         = (real*) root_fftw::fft_malloc(sizeof(real) * total_local_size);                         // array storing data
-   real   *SendBuf      = new real [ amr->NPatchComma[0][1]*CUBE(PS1) ];         // MPI send buffer for data
-   real   *RecvBuf      = new real [ NX0_TOT[0]*NX0_TOT[1]*NRecvSlice ];         // MPI recv buffer for data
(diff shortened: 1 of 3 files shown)


src/Output/Output_BasePowerSpectrum.cpp
```diff
@@ -58,6 +58,11 @@ void Output_BasePowerSpectrum( const char *FileName, const long TVar )
 #  endif // #  if ( SUPPORT_FFTW == FFTW3 ) ... # else
 #  endif // #  ifdef SERIAL ... # else
 
+// check integer overflow (in a non-rigorous way)
+   if ( total_local_size <= 0 )
```
KunkelAlexander, line 62: For which platforms and compilers does this work? According to https://users.cs.utah.edu/~regehr/papers/overflow12.pdf the overflow behaviour for signed integers should be undefined in C. 
For unsigned types, the behaviour seems to be to wrap around. Could we make mpi_index_int an unsigned integer for FFTW2 and then check whether `(long) local_nx*local_ny*local_nz > mpi_index_int`to detect an overflow more reliably?

src/Output/Output_BasePowerSpectrum.cpp
```diff
@@ -76,11 +81,11 @@ void Output_BasePowerSpectrum( const char *FileName, const long TVar )
...
    double *PS_total     = NULL;
-   real   *VarK         = (real*) root_fftw::fft_malloc(sizeof(real) * total_local_size);                         // array storing data
-   real   *SendBuf      = new real [ amr->NPatchComma[0][1]*CUBE(PS1) ];         // MPI send buffer for data
-   real   *RecvBuf      = new real [ NX0_TOT[0]*NX0_TOT[1]*NRecvSlice ];         // MPI recv buffer for data
-   long   *SendBuf_SIdx = new long [ amr->NPatchComma[0][1]*PS1 ];               // MPI send buffer for 1D coordinate in slab
-   long   *RecvBuf_SIdx = new long [ NX0_TOT[0]*NX0_TOT[1]*NRecvSlice/SQR(PS1) ];// MPI recv buffer for 1D coordinate in slab
+   real   *VarK         = (real*) root_fftw::fft_malloc(sizeof(real) * total_local_size); // array storing data
+   real   *SendBuf      = new real [ (long)amr->NPatchComma[0][1]*CUBE(PS1) ];            // MPI send buffer for data
+   real   *RecvBuf      = new real [ (long)NX0_TOT[0]*NX0_TOT[1]*NRecvSlice ];            // MPI recv buffer for data
+   long   *SendBuf_SIdx = new long [ amr->NPatchComma[0][1]*PS1 ];                        // MPI send buffer for 1D coordinate in slab
```
KunkelAlexander, line 87: Maybe we can add a (long) conversion for `RecfBuf `too?

src/SelfGravity/CPU_Poisson/CPU_PoissonSolver_FFT.cpp
```diff
@@ -205,6 +205,12 @@ void CPU_PoissonSolver_FFT( const real Poi_Coeff, const int SaveSg, const double
 #  endif // #  if ( SUPPORT_FFTW == FFTW3 ) ... # else
 #  endif
 
+// check integer overflow (in a non-rigorous way)
+   if ( total_local_size <= 0 )
```
KunkelAlexander, line 209: Same as above.

src/SelfGravity/CPU_Poisson/CPU_PoissonSolver_FFT.cpp
```diff
@@ -226,7 +232,7 @@ void CPU_PoissonSolver_FFT( const real Poi_Coeff, const int SaveSg, const double
    real *SendBuf      = new real [ (long)amr->NPatchComma[0][1]*CUBE(PS1) ];          // MPI send buffer for density and potential
    real *RecvBuf      = new real [ (long)NX0_TOT[0]*NX0_TOT[1]*NRecvSlice ];          // MPI recv buffer for density and potentia
    long *SendBuf_SIdx = new long [ amr->NPatchComma[0][1]*PS1 ];                      // MPI send buffer for 1D coordinate in slab
```
KunkelAlexander, line 234: Same as above.

src/SelfGravity/Init_GreenFuncK.cpp
```diff
@@ -49,6 +49,12 @@ void Init_GreenFuncK()
 #  endif // #  if ( SUPPORT_FFTW == FFTW3 ) ... # else
 #  endif // # ifdef SERIAL
 
+// check integer overflow (in a non-rigorous way)
+   if ( total_local_size <= 0 )
```
KunkelAlexander, line 53: Same as above.

KunkelAlexander (commented):
Thanks for finding this bug! I was frustrated when the N = 2048 simulations always failed and gave up thinking that the RAM was not enough.

src/Output/Output_BasePowerSpectrum.cpp
KunkelAlexander: Based on [this](https://stackoverflow.com/questions/3944505/detecting-signed-overflow-in-c-c) link, I suspect your solution might still not be safe as behaviour after an overflow is undefined according to the C standard. They suggest using a type that is large enough to hold the result and then checking against INT_MIN and INT_MAX. I believe this solution should be robust. 

```
assert(LLONG_MAX>INT_MAX);
long long size_long = (long long)local_nx*(long long)local_ny*(long long)local_nz;
if (size_long < INT_MIN || size_long > INT_MAX) 
    Aux_Error( ERROR_INFO, "size_long (%ld) != size_int (%ld) and/or total_local_size (%ld) <= 0 for FFT\n"
                   "        --> Likely due to integer overflow. Consider using more MPI processes !!\n",
               size_long, (long)size_int, (long)total_local_size );
```

JeiwHuang (approved):

PR #233, opened 2023-09-02
Use heap instead of stack memory for large arrays in Refine()

hyschive:
Use heap instead of stack memory for large arrays in `Refine()`. It is important for larger `PATCH_SIZE`. For example, the original code with `--gravity=true --mhd=true --patch_size=32 --double=true` will crash due to stack overflow in `Refine()` for both serial and parallel runs.

Thank @koarakawaii for reporting this issue!


koarakawaii (approved):
Looks good and the `gamer` thus compiled can be run successfully!

PR #286, opened 2024-02-04
Generalize Aux_ComputeProfile()

hyschive:
This PR generalizes the radial profile routine `Aux_ComputeProfile()`:
- Support more fields by calling `Prepare_PatchData()`, such as particle density and cell-centered magnetic fields
- Support temporal interpolation even when disabling `OPT__INT_TIME`
- When enabling temporal interpolation, the radial velocity profile (`_VELR`) will be slightly different from the previous results due to a different weighting approach

### DIFF
--- a/src/Auxiliary/Aux_ComputeProfile.cpp
+++ b/src/Auxiliary/Aux_ComputeProfile.cpp
@@ -253,24 +253,6 @@ void Aux_ComputeProfile( Profile_t *Prof[], const double Center[], const double 
    OPT__INT_TIME = ( PrepTimeIn >= 0.0 ) ? true : false;
 
 
-// set undefined fields to _NONE for simplicity
-#  ifndef _VELX
-#  define _VELX _NONE
-#  endif
-#  ifndef _VELY
-#  define _VELY _NONE
-#  endif
-#  ifndef _VELZ
-#  define _VELZ _NONE
-#  endif
-#  ifndef _VELR
-#  define _VELR _NONE
-#  endif
-#  ifndef _POTE
-#  define _POTE _NONE
-#  endif
-
-
 // loop over all target levels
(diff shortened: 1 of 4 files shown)


src/Auxiliary/Aux_ComputeProfile.cpp
```diff
@@ -120,28 +128,42 @@ void Aux_ComputeProfile( Profile_t *Prof[], const double Center[], const double
    if ( PatchType != PATCH_LEAF  &&  PatchType != PATCH_NONLEAF  &&
         PatchType != PATCH_BOTH  &&  PatchType != PATCH_LEAF_PLUS_MAXNONLEAF )
       Aux_Error( ERROR_INFO, "incorrect PatchType (%d) !!\n", PatchType );
-#  endif
 
+#  ifdef OPENMP
+   if (  omp_in_parallel()  &&  ( (PrepTimeIn >= 0.0 && !OPT__INT_TIME) ||
```
hfhsieh, line 133: This check prevents users from applying temporal interpolation after computing the profiles (e.g., the GREP profiles in the `ccsn` branch). How about printing a warning message instead? Or solve the thread-unsafe issue (`OPT__INT_TIME`)?

hfhsieh (commented):
Looks good to me. I left a comment regarding the check of `OPT__INT_TIME`.

src/Auxiliary/Aux_ComputeProfile.cpp
hyschive: @hfhsieh I'm confused. In the `ccsn` branch, where we do launch multiple threads to _call_ this function in parallel? I thought we only apply OpenMP _inside_ this function.

hfhsieh (approved):

PR #287, opened 2024-02-05
Hydro/ClusterMerger: correct the attribute to write B_IC file

Xian-Hsu-2023:
The attribute bfield.wrtie_to_h5() is no longer existing in the package cluster_generator. Instead, it inherits write_file() function from the class ClusterField now.


hyschive:
@jzuhone Could you take a look?

PR #288, opened 2024-02-22
Set VelR=0 when r==0 in Aux_ComputeProfile()

hyschive:
Set the radial velocity to zero for the innermost bin if the profile center coincides with a cell center


hfhsieh (approved):
Looks good.

PR #296, opened 2024-04-09
Generalize Par_SortByPos()

### LINKED ISSUE #292: Generalize Par_SortByPos()
- [ ] Add templates `template <typename T>` for different data types of `PosX/Y/Z`
- [ ] Replace `const long NPar, const real_par *PosX, const real_par *PosY, const real_par *PosZ` with `const int NField, const long FieldSize, const T **Array`, where `Array` has the dimension `[NField][FieldSize]`.
- [ ] Rename it as `Mis_SortByMultiField()` to support non-particle data.
- [ ] Print a warning message for degenerate data.
- [ ] Remove the comment `//  2. Currently IdxTable has the type "int" instead of "long" since Mis_Heapsort() doesn't support "long" yet...`
- [ ] In `FB_AdvanceDt.cpp`, declare `PCr` as `int **` instead of `real_par **`.

ChunYen-Chen:
As the title, this PR resolves #292.


src/Miscellaneous/Mis_SortByMultiField.cpp
hsinhaoHHuang, line 1: - How about not using `IdxTable_copy` and updating `IdxTable` directly?
- Is it possible to support user-specified orders (by an array of indices) for sorting fields?
- How about making the recursion more symmetric by only sending the needed  `IdxTable` with size `NSort` into the next recursion?
    - Then, we have to move the initial IdxTable outside this function.

hsinhaoHHuang (changes requested):
I have left some comments. Please take a look. Thanks!
Also, could you please merge with the main branch first so that I can compare the results more clearly?

ChunYen-Chen:
@hsinhaoHHuang The algorithm is updated as reviewed comments. Please help to review the latest version. Thanks!

hsinhaoHHuang (changes requested):
@ChunYen-Chen Thank you for the update.

I have tested it, and I think it works well.
The sorted results are the same in the Plummer test problem with feedback.

I have only some minor comments now.

src/Miscellaneous/Mis_SortByMultiField.cpp
```diff
@@ -0,0 +1,84 @@
...
+//                SortOrder : The order of sorting field has a size of NOrder.
+//                NOrder    : The size of the sort field order.
+//
+// Return      :  IdxTable
+//-------------------------------------------------------------------------------------------------------
+template <typename T>
+void Mis_SortByMultiField( T **Array, long *IdxTable, const long NSort, const int *SortOrder, const int NOrder )
+{
+   if ( NSort  < 0 )   Aux_Error( ERROR_INFO, "NSort must be greater than zero!!\n" );
+   if ( NOrder < 1 )   Aux_Error( ERROR_INFO, "NOrder must be greater than one!!\n" );
```
hsinhaoHHuang, line 23: "greater than or equal to"

hsinhaoHHuang (approved):

ChunYen-Chen:
@hsinhaoHHuang Thanks for the detailed comment. I have updated the description of the parameters and changed the function name.

hsinhaoHHuang (approved):
Great! I have also tested it again. I don't find any further issues. I think everything works well now.

src/Miscellaneous/Mis_SortByRows.cpp
```diff
@@ -0,0 +1,98 @@
...
+//                NSort     : Number of columns to be sorted in the array must less than or equal to n.
+//                SortOrder : Array with a size of NOrder specifies which rows to compare first, second, etc.
+//                NOrder    : Size of SortOrder must be less than or equal to m.
+//
+// Return      :  IdxTable
+//-------------------------------------------------------------------------------------------------------
+template <typename T>
+void Mis_SortByRows( T **Array, long *IdxTable, const long NSort, const int *SortOrder, const int NOrder )
+{
+   if ( NSort  < 0 )   Aux_Error( ERROR_INFO, "NSort must be greater than or equal to zero!!\n" );
```
hyschive, line 36: Simplify the error message as `"NSort < 0 !!\n"`

src/Miscellaneous/Mis_SortByRows.cpp
```diff
@@ -0,0 +1,98 @@
...
+   if ( NOrder < 1 )   Aux_Error( ERROR_INFO, "NOrder must be greater than or equal to one!!\n" );
```
hyschive, line 37: Simplify the error message as `"NOrder < 1 !!\n"`

src/Miscellaneous/Mis_SortByRows.cpp
```diff
@@ -0,0 +1,98 @@
...
+// 3. check the same value
+   for (long i=0; i<NSort-1; i++)
+   {
+      long NSameVal = 1; // 1 --> itself
+
+      while ( i+NSameVal < NSort  &&  Array_Sorted[i] == Array_Sorted[i+NSameVal] )   NSameVal++;
+
+      if ( NSameVal == 1 )   continue;
+
+      if ( NOrder == 1 )
```
hyschive, line 65: Change to `else if ( NOrder == 1 )` to improve clarity.

hyschive (commented):
@ChunYen-Chen Looks great! I have added some comments. The only major comment is to add a new API routine so that users do not have to initialize `IdxTable[]` by themselves.

ChunYen-Chen:
@hyschive Thanks for your review! I have modified the code according to the comments.

hyschive (approved):

PR #310, opened 2024-05-20
Minimum/maximum angular resolution

### LINKED ISSUE #306: Minimum/maximum angular resolution
Port the PR hyschive/gamer-fork#117 to the `main` branch. Regard it as a general-purpose refinement criterion independent of GREP and CCSN.

ChunYen-Chen:
Resolve #306.

A new refinement criterion from https://github.com/hyschive/gamer-fork/pull/117.


src/Refine/Flag_Check.cpp
```diff
@@ -338,3 +353,60 @@ bool Check_Curl( const int i, const int j, const int k,
...
+//                CenX        : x coordinate of center of calculating angular resolution
+//                CenY        : y coordinate of center of calculating angular resolution
+//                CenZ        : z coordinate of center of calculating angular resolution
+//                AngRes_Max  : the maximum allowed angular resolution
+//                AngRes_Min  : the minimum allowed angular resolution
+//                Refine_Rad  : the radius of must refined at level (lv)
+//
+// Return      :  0 : if the cell is not in the refine region.
+//                1 : if the minimum angular resolution is     reached or  r <  Refine_Rad
+//                2 : if the minimum angular resolution is not reached and r >= Refine_Rad
```
hfhsieh, line 378: - remove the trailing dot in this line and line 376
- should we capitalize the first "the" in the Parameter section, given that both `the` and `The` are used in GAMER?

src/Refine/Flag_Check.cpp
```diff
@@ -338,3 +353,60 @@ bool Check_Curl( const int i, const int j, const int k,
...
+   if ( AngRes_Max > 0.0  &&  2.0 * R * AngRes_Max > dh )
+      return 0;
+
+// (2) refine the region within r < Refine_Rad and the innermost cells
+   if ( R < dh  ||  R * UNIT_L < Refine_Rad )
+      return 1;
+
+// (3) check if the minimum angular resolution is reached
+   if ( AngRes_Min > 0.0  &&  R * AngRes_Min < dh )
+      return 1;
```
hfhsieh, line 408: - I think the Criterion (2) should be listed first if the priority of `Refine_Rad` is higher than the angular resolution. 
- It seems that we assume the input `Refine_Rad` is in some units that are not mentioned in `Input__Flag_AngularResolution`.  Should we force the unit of `Refine_Rad` to be code unit, or convert its value to in code unit, in order to save the time for unit conversion (`R * UNIT_L < Refine_Rad`)?
- In the document we should note the users that the priority of `AngRes_Max` is higher than `AngRes_Min` in case of `2.0 * AngRes_Max > AngRes_Min`.

src/Init/Init_Load_FlagCriteria.cpp
```diff
@@ -195,6 +198,10 @@ void Init_Load_FlagCriteria()
 //          OPT__FLAG_NPAR_PATCH/CELL load integers
             else if ( FlagMode == 6 )  sscanf( input_line, "%d%d",  &Trash, &FlagTable_NParPatch[lv] );
             else if ( FlagMode == 7 )  sscanf( input_line, "%d%d",  &Trash, &FlagTable_NParCell [lv] );
+//          OPT__FLAG_ANGULAR has four columns to be loaded
+            else if ( FlagMode == 14 ) sscanf( input_line, "%d%lf%lf%lf", &Trash, &FlagTable_Angular[lv][0],
+                                                                                  &FlagTable_Angular[lv][1],
+                                                                                  &FlagTable_Angular[lv][2] );
```
hfhsieh, line 204: Should we check whether `AngRes_Max < AngRes_Min` if both are given, and raise a warning/error here or somewhere?

src/Refine/Flag_Check.cpp
```diff
@@ -338,3 +353,60 @@ bool Check_Curl( const int i, const int j, const int k,
...
+//                AngRes_Min  : the minimum allowed angular resolution
+//                Refine_Rad  : the radius of must refined at level (lv)
+//
+// Return      :  0 : if the cell is not in the refine region.
+//                1 : if the minimum angular resolution is     reached or  r <  Refine_Rad
+//                2 : if the minimum angular resolution is not reached and r >= Refine_Rad
+//-------------------------------------------------------------------------------------------------------
+int Check_Angular( const int i, const int j, const int k, const int lv, const int PID,
+                   const real CenX, const real CenY, const real CenZ,
+                   const real AngRes_Max, const real AngRes_Min, const real Refine_Rad )
```
hfhsieh, line 382: The datatype of `CenX/Y/Z`, `AngRes_Max/Min`, and `Refine_Rad` should be `double`

hfhsieh (commented):
This PR looks fine to me. However, I have some fundamental questions about this feature. Perhaps you have already discussed these.
1. Do we need different `AngRes_Max/Min` values at different levels?
2. Should we support a feature that the reference center to vary over time?
3. The unit of the input AngRes_Max/Min is implicitly assumed to be radians. Should we support a feature that allows users to choose between radians and degrees?
4. This PR contains two features for mesh refinement: angular resolution and `Refine_Rad` (always refine cells within a certain radius). Both features are controlled by the runtime parameter `OPT__FLAG_ANGULAR`. Should we separate them?

One minor issue unrelated to this PR:
- Should we unify the labels representing the reference center? `CEN` is used in this PR and `RECORD_CENTER`,  and `ORIGIN` is used for recording the angular momentum?

src/Refine/Flag_Check.cpp
hyschive: `Refine_Rad` should be in the code unit like most of the built-in parameters. Remove `UNIT_L`.

hyschive:
> 1. Do we need different `AngRes_Max/Min` values at different levels?

Sometimes we may want to apply these criteria only above or below certain AMR levels. We may also want to increase the angular resolution above certain levels. So I suggest keeping it. What do you two think?

> 2. Should we support a feature that the reference center to vary over time?

Users can achieve that by modifying `ANGULAR_CEN_X/Y/Z` during runtime. I've created a corresponding feature request (#322).

> 3. The unit of the input AngRes_Max/Min is implicitly assumed to be radians. Should we support a feature that allows users to choose between radians and degrees?

I think it's OK to stick with just one unit system. @hfhsieh Do you prefer radians or degrees?

> 4. This PR contains two features for mesh refinement: angular resolution and `Refine_Rad` (always refine cells within a certain radius). Both features are controlled by the runtime parameter `OPT__FLAG_ANGULAR`. Should we separate them?

Yes, let's separate them. Here are two related suggestions.
- We should allow the two refinement criteria to have different centers.
- How about adding a parameter to control the maximum level for `Refine_Rad` (and setting its default to `MAX_LEVEL`)?

> * Should we unify the labels representing the reference center? `CEN` is used in this PR and `RECORD_CENTER`,  and `ORIGIN` is used for recording the angular momentum?

I think it's OK to have different names since they actually mean different things. To me, the center refers to the central point of a given object (e.g., a halo or PNS), which can change with time, while the origin is for a coordinate system, which is usually fixed during a simulation.

hfhsieh:
> > 1. Do we need different `AngRes_Max/Min` values at different levels?
> 
> Sometimes we may want to apply these criteria only above or below certain AMR levels. We may also want to increase the angular resolution above certain levels. So I suggest keeping it. What do you two think?

Agreed. Then, should we need one more function for checking the input values in `Input__Flag_XXX` loaded in `Init_Load_FlagCriteria()` (or it already exists)? 

> > 2. Should we support a feature that the reference center to vary over time?
> 
> Users can achieve that by modifying `ANGULAR_CEN_X/Y/Z` during runtime. I've created a corresponding feature request (#322).
> 

Once this new feature is supported, should we support some built-in definitions of the centers (like box center, cell with minimum potential, CoM, etc).

> > 3. The unit of the input AngRes_Max/Min is implicitly assumed to be radians. Should we support a feature that allows users to choose between radians and degrees?
> 
> I think it's OK to stick with just one unit system. @hfhsieh Do you prefer radians or degrees?
> 

OK. Then let's stick with radians.

> > 4. This PR contains two features for mesh refinement: angular resolution and `Refine_Rad` (always refine cells within a certain radius). Both features are controlled by the runtime parameter `OPT__FLAG_ANGULAR`. Should we separate them?
> 
> Yes, let's separate them. Here are two related suggestions.
> 
>     * We should allow the two refinement criteria to have different centers.
> 
>     * How about adding a parameter to control the maximum level for `Refine_Rad` (and setting its default to `MAX_LEVEL`)?
> 
> 
On the maximum level for `Refine_Rad`, I think it can be controlled by the value specified in `Input__Flag_XXX`. Nevertheless, adding a runtime parameter is useful for correcting the value in `Input__Flag_XXX`, or in the case that we specify `Refine_Rad` via `Input__Parameter`.

> > * Should we unify the labels representing the reference center? `CEN` is used in this PR and `RECORD_CENTER`,  and `ORIGIN` is used for recording the angular momentum?
> 
> I think it's OK to have different names since they actually mean different things. To me, the center refers to the central point of a given object (e.g., a halo or PNS), which can change with time, while the origin is for a coordinate system, which is usually fixed during a simulation.

Got it.

hyschive:
> Agreed. Then, should we need one more function for checking the input values in Input__Flag_XXX loaded in Init_Load_FlagCriteria() (or it already exists)?

Yes, let's move `Refine_Rad` to a separate `Input__Flag_*` table.

> Once this new feature is supported, should we support some built-in definitions of the centers (like box center, cell with minimum potential, CoM, etc).

I tend to leave it as a future work. For CCSN simulations, we can just set the refinement center to `GREP_Center` in `Flag_UserWorkBeforeFlag_Ptr`.

> On the maximum level for Refine_Rad, I think it can be controlled by the value specified in Input__Flag_XXX. Nevertheless, adding a runtime parameter is useful for correcting the value in Input__Flag_XXX, or in the case that we specify Refine_Rad via Input__Parameter.

Note that the check `if ( R < dh )` will force refinement even with an extremely small `Refine_Rad`. So how about disabling this refinement criterion when `Refine_Rad < 0.0`?

hfhsieh:
> > Once this new feature is supported, should we support some built-in definitions of the centers (like box center, cell with minimum potential, CoM, etc).
> 
> I tend to leave it as a future work. For CCSN simulations, we can just set the refinement center to `GREP_Center` in `Flag_UserWorkBeforeFlag_Ptr`.
> 
Agreed. Also, it might be complicated to determine when and how to update the center (for other cases) since this function `Flag_UserWorkBeforeFlag_Ptr()` is invoked during sub-steps in the current design.

> > On the maximum level for Refine_Rad, I think it can be controlled by the value specified in Input__Flag_XXX. Nevertheless, adding a runtime parameter is useful for correcting the value in Input__Flag_XXX, or in the case that we specify Refine_Rad via Input__Parameter.
> 
> Note that the check `if ( R < dh )` will force refinement even with an extremely small `Refine_Rad`. So how about disabling this refinement criterion when `Refine_Rad < 0.0`?

Yes, I missed in the review.

hyschive:
> Agreed. Also, it might be complicated to determine when and how to update the center (for other cases) since this function Flag_UserWorkBeforeFlag_Ptr() is invoked during sub-steps in the current design.

I'm a bit confused by your comment. `Flag_UserWorkBeforeFlag_Ptr()` is always invoked just before refinement so that users do not need to worry about _when_ to update the center. And it's totally up to you as to _how_ to update the center.

hfhsieh:
> I'm a bit confused by your comment. `Flag_UserWorkBeforeFlag_Ptr()` is always invoked just before refinement so that users do not need to worry about _when_ to update the center. And it's totally up to you as to _how_ to update the center.

One case is that at the final sub-step on `level=lv` (`Time[lv] == Time[lv-1]`), the center is updated in `Flag_UserWorkBeforeFlag_Ptr()` on `level=lv` so that we can skip the update of center on `level=lv-1` to improve performance.

hyschive:
> One case is that at the final sub-step on `level=lv` (`Time[lv] == Time[lv-1]`), the center is updated in `Flag_UserWorkBeforeFlag_Ptr()` on `level=lv` so that we can skip the update of center on `level=lv-1` to improve performance.

Indeed there can be some redundant operations when refining multiple levels at once. But let's not worry about that for now.

src/Refine/Flag_Check.cpp
hfhsieh: It appears to be a bug in the `ccsn` branch that we overlooked.

src/Init/Init_Load_Parameter.cpp
```diff
@@ -183,6 +183,10 @@ void Init_Load_Parameter()
    ReadPara->Add( "ANGULAR_CEN_X",              &ANGULAR_CEN_X,                  -1.0,             NoMin_double,  NoMax_double   );
    ReadPara->Add( "ANGULAR_CEN_Y",              &ANGULAR_CEN_Y,                  -1.0,             NoMin_double,  NoMax_double   );
    ReadPara->Add( "ANGULAR_CEN_Z",              &ANGULAR_CEN_Z,                  -1.0,             NoMin_double,  NoMax_double   );
+   ReadPara->Add( "OPT__FLAG_RADIAL",          &OPT__FLAG_RADIAL,               false,             Useless_bool,  Useless_bool   );
+   ReadPara->Add( "RADIAL_CEN_X",              &RADIAL_CEN_X,                    -1.0,             NoMin_double,  NoMax_double   );
+   ReadPara->Add( "RADIAL_CEN_Y",              &RADIAL_CEN_Y,                    -1.0,             NoMin_double,  NoMax_double   );
+   ReadPara->Add( "RADIAL_CEN_Z",              &RADIAL_CEN_Z,                    -1.0,             NoMin_double,  NoMax_double   );
```
hfhsieh, line 189: These `&PARAM` and the `false` are misaligned.

src/Init/Init_Load_FlagCriteria.cpp
```diff
@@ -140,76 +144,78 @@ void Init_Load_FlagCriteria()
...
+         }
+//       OPT__FLAG_NPAR_PATCH/CELL load integers
+         else if ( FlagMode == 6 )  sscanf( input_line, "%d%d",  &Trash, &FlagTable_NParPatch[lv] );
+         else if ( FlagMode == 7 )  sscanf( input_line, "%d%d",  &Trash, &FlagTable_NParCell [lv] );
+//       OPT__FLAG_ANGULAR has three columns to be loaded
+         else if ( FlagMode == 14 )
+         {
+            sscanf( input_line, "%d%lf%lf", &Trash, &FlagTable_Angular[lv][0], &FlagTable_Angular[lv][1] );
+            if ( FlagTable_Angular[lv][0] < FlagTable_Angular[lv][1] )
+               Aux_Message( stderr, "WARNING : AngRes_Max < AngRes_Min at lv %d\n", lv );
```
hfhsieh, line 210: This check is applied only when both `AngRes_Min` and `AngRes_Max` are set, and check that `AngRes_Max` is smaller than `AngRes_Min`. Thus, the condition should be 
~~~ c
if ( FlagTable_Angular[lv][0] > 0.0   &&
     FlagTable_Angular[lv][1] > 0.0   &&
     FlagTable_Angular[lv][0] > FlagTable_Angular[lv][1] )
~~~

The warning message should be (also provide more information)
~~~ c
Aux_Message( stderr, "WARNING : AngRes_Max (%20.14e) > AngRes_Min (%20.14e) at lv %d\n", 
             FlagTable_Angular[lv][0] , FlagTable_Angular[lv][1], lv );
~~~

src/Refine/Flag_Check.cpp
```diff
@@ -362,26 +372,25 @@ bool Check_Curl( const int i, const int j, const int k,
 //                than the maximum angular resolution. Also, check if the cell is in Refined_Rad.
```
hfhsieh, line 372: - This description `Also, check if the cell is in Refined_Rad.` is obsolete
- Remove the trailing dot. Also in line 423 (the description in `Check_Radial()`)

src/Refine/Flag_Check.cpp
```diff
@@ -59,19 +61,27 @@ bool Flag_Check( const int lv, const int PID, const int i, const int j, const in
...
       if ( Flag )              return Flag;
    }
 
 
+// check radial resolution
+// ===========================================================================================
+   if ( OPT__FLAG_RADIAL )
+   {
+      Flag |= Check_Radial( i, j, k, lv, PID, RADIAL_CEN_X, RADIAL_CEN_Y, RADIAL_CEN_Z, FlagTable_Radial[lv] );
+      if ( Flag )              return Flag;
```
hfhsieh, line 81: Remove the redundant whitespaces before `return`

hfhsieh (commented):
The revision looks good to me. I have left some comments. 

We may need to switch the checks `if ( OPT__FLAG_ANGULAR )` and `if ( OPT__FLAG_RADIAL )` in `Flag_Check()` if we want the priority of radial resolution is higher than angular resolution.

hfhsieh (approved):
Looks good to me.

I was supposed that you use `FormatVersion = 2480` because the index `2479` is used in  PR #304. However, this is fine.

ChunYen-Chen:
@hfhsieh Thanks for reviewing!

doc/wiki/Runtime-Parameters:-Refinement.md
hyschive, line 1: Add `ANGULAR_CEN_X/Y/Z` and `RADIAL_CEN_X/Y/Z`.

doc/wiki/Runtime-Parameters:-Refinement.md
```diff
@@ -363,6 +365,26 @@ any refinement. Instead, it simply forbids refinement outside the
...
 
+<a name="OPT__FLAG_ANGULAR"></a>
+* #### `OPT__FLAG_ANGULAR` &ensp; (0=off, 1=on) &ensp; [0]
+    * **Description:**
+Refinement criterion: Cell radian from the given centre (`ANGULAR_CEN_X`,
+`ANGULAR_CEN_Y`, `ANGULAR_CEN_Z`). Specify the refinement thresholds
+on different levels in the input file `Input__Flag_AngularResolution`
+with the [[specific format | Runtime Parameters#input__flag_]].
+An example file can be found at `example/input/Input__Flag_AngularResolution`.
+    * **Restriction:**
```
hyschive, line 376:
```suggestion
Refinement criterion: angular resolution with respect to the specified center
(`ANGULAR_CEN_X`, `ANGULAR_CEN_Y`, `ANGULAR_CEN_Z`).
Cells located at a distance greater than `AngRes_Max_R` are not allowed to exceed
the angular resolution `AngRes_Max`. Cells are refined if their angular resolution
is lower than `AngRes_Min`. Set `AngRes_Max < 0.0` or `AngRes_Min < 0.0` to
disable the respective criterian.
Specify the refinement thresholds `AngRes_Max, AngRes_Min, AngRes_Max_R`
on different levels in the input file `Input__Flag_AngularResolution`
with the [[specific format | Runtime Parameters#input__flag_]].
An example file can be found at `example/input/Input__Flag_AngularResolution`.
    * **Restriction:**
`AngRes_Max` has higher priority over `AngRes_Min`.
It is generally recommended to set `AngRes_Max < 0.5*AngRes_Min`. 
```

doc/wiki/Runtime-Parameters:-Refinement.md
```diff
@@ -363,6 +365,26 @@ any refinement. Instead, it simply forbids refinement outside the
...
+
+<a name="OPT__FLAG_RADIAL"></a>
+* #### `OPT__FLAG_RADIAL` &ensp; (0=off, 1=on) &ensp; [0]
+    * **Description:**
+Refinement criterion: Distance to the given centre (`RADIAL_CEN_X`,
+`RADIAL_CEN_Y`, `RADIAL_CEN_Z`). Specify the refinement thresholds
+on different levels in the input file `Input__Flag_RadialResolution`
+with the [[specific format | Runtime Parameters#input__flag_]].
+An example file can be found at `example/input/Input__Flag_RadialResolution`.
```
hyschive, line 385:
```suggestion
Refinement criterion: distance to the specified center (`RADIAL_CEN_X`,
`RADIAL_CEN_Y`, `RADIAL_CEN_Z`).
Cells are refined if they are located at a distance smaller than `Refine_Rad`.
The value of `Refine_Rad` can be specified in the input file `Input__Flag_RadialResolution`
with the [[specific format | Runtime Parameters#input__flag_]].
An example file can be found at `example/input/Input__Flag_RadialResolution`.
```

hyschive (changes requested):
@ChunYen-Chen @hfhsieh Thanks for the contribution! I’ve added a few comments. The most significant one concerns separating `Check_Angular()` and ensuring that `Check_Radial()` is called after `OPT__FLAG_INTERFERENCE`.

example/test_problem/Template/Input__Parameter
```diff
@@ -167,13 +167,13 @@ OPT__FLAG_USER                0           # flag: user-defined (Input__Flag_User
 OPT__FLAG_USER_NUM            1           # number of threshold values in user-defined table (Input__Flag_User) [1]
 OPT__FLAG_REGION              0           # flag: specify the regions **allowed** to be refined -> edit "Flag_Region.cpp" [0]
 OPT__FLAG_ANGULAR             0           # flag: angular resolution (Input__Flag_AngularResolution) [0]
-ANGULAR_CEN_X                -1.0         # x coordinate of angular resolution center (if one of ANGULAR_CEN_X/Y/Z < 0 -> box center) [-1.0]
-ANGULAR_CEN_Y                -1.0         # y coordinate of angular resolution center (if one of ANGULAR_CEN_X/Y/Z < 0 -> box center) [-1.0]
-ANGULAR_CEN_Z                -1.0         # z coordinate of angular resolution center (if one of ANGULAR_CEN_X/Y/Z < 0 -> box center) [-1.0]
+ANGULAR_CEN_X                -1.0         # x center coordinate for OPT__FLAG_ANGULAR (< 0=auto -> box center) [-1.0]
+ANGULAR_CEN_Y                -1.0         # y center coordinate for OPT__FLAG_ANGULAR (< 0=auto -> box center) [-1.0]
+ANGULAR_CEN_Z                -1.0         # z center coordinate for OPT__FLAG_ANGULAR (< 0=auto -> box center) [-1.0]
```
hfhsieh, line 172: These runtime parameters were renamed to `FLAG_ANGULAR_CEN_[X/Y/Z]`. Same for `RADIAL_CEN_[X/Y/Z]`

doc/wiki/Runtime-Parameters-related/Runtime-Parameters:-Refinement.md
```diff
@@ -368,23 +374,69 @@ specified regions.
 <a name="OPT__FLAG_ANGULAR"></a>
 * #### `OPT__FLAG_ANGULAR` &ensp; (0=off, 1=on) &ensp; [0]
     * **Description:**
-Refinement criterion: Cell radian from the given centre (`ANGULAR_CEN_X`,
-`ANGULAR_CEN_Y`, `ANGULAR_CEN_Z`). Specify the refinement thresholds
+Refinement criterion: angular resolution with respect to the specified center
+([FLAG_ANGULAR_CEN_X](#FLAG_ANGULAR_CEN_X), [FLAG_ANGULAR_CEN_Y](#FLAG_ANGULAR_CEN_Y),
+[FLAG_ANGULAR_CEN_Z](#FLAG_ANGULAR_CEN_Z)). Cells located at a distance greater than
+`AngRes_Max_R` are not allowed to exceed the angular resolution `AngRes_Max`.
+Cells are refined if their angular resolution is lower than `AngRes_Min`. Set
```
hyschive, line 381:
```suggestion
[FLAG_ANGULAR_CEN_Z](#FLAG_ANGULAR_CEN_Z)). Cells located at a distance greater than
`AngRes_Max_R` are not allowed to exceed the angular resolution `AngRes_Max` (in radians).
Cells are refined if their angular resolution is lower than `AngRes_Min` (in radians). Set
```

src/Refine/Flag_Check.cpp
```diff
@@ -416,36 +430,30 @@ bool Check_Curl( const int i, const int j, const int k,
...
 //                lv           : Refinement level of the target patch
 //                PID          : ID of the target patch
-//                CenX         : x coordinate of center of calculating angular resolution
-//                CenY         : y coordinate of center of calculating angular resolution
-//                CenZ         : z coordinate of center of calculating angular resolution
-//                AngRes_Max   : The maximum allowed angular resolution
-//                AngRes_Min   : The minimum allowed angular resolution
-//                AngRes_Max_R : The minimum radius to apply AngRes_Max
+//                CenX/Y/Z     : x/y/z-coordinate of the center for calculating angular resolution
+//                AngRes_Max   : Maximum allowed angular resolution
```
hyschive, line 442:
```suggestion
//                AngRes_Max   : Maximum allowed angular resolution (in radians)
```

src/Refine/Flag_Check.cpp
```diff
@@ -454,32 +462,61 @@ int Check_Angular( const int i, const int j, const int k, const int lv, const in
...
+// Function    :  Check_Angular_Min
+// Description :  Check if dh/R at cell (i,j,k) is larger than the minimum angular resolution
+//
+// Note        :  1. Enabled by the runtime option "OPT__FLAG_ANGULAR"
+//
+// Parameter   :  i,j,k        : Target cell indices in the patch amr->patch[0][lv][PID]
+//                lv           : Refinement level of the target patch
+//                PID          : ID of the target patch
+//                CenX/Y/Z     : x/y/z-coordinate of the center for calculating angular resolution
+//                AngRes_Min   : Minimum allowed angular resolution
```
hyschive, line 481:
```suggestion
//                AngRes_Min   : Minimum allowed angular resolution (in radians)
```

example/test_problem/Template/Input__Parameter
```diff
@@ -167,13 +167,13 @@ OPT__FLAG_USER                0           # flag: user-defined (Input__Flag_User
 OPT__FLAG_USER_NUM            1           # number of threshold values in user-defined table (Input__Flag_User) [1]
 OPT__FLAG_REGION              0           # flag: specify the regions **allowed** to be refined -> edit "Flag_Region.cpp" [0]
 OPT__FLAG_ANGULAR             0           # flag: angular resolution (Input__Flag_AngularResolution) [0]
-ANGULAR_CEN_X                -1.0         # x coordinate of angular resolution center (if one of ANGULAR_CEN_X/Y/Z < 0 -> box center) [-1.0]
-ANGULAR_CEN_Y                -1.0         # y coordinate of angular resolution center (if one of ANGULAR_CEN_X/Y/Z < 0 -> box center) [-1.0]
-ANGULAR_CEN_Z                -1.0         # z coordinate of angular resolution center (if one of ANGULAR_CEN_X/Y/Z < 0 -> box center) [-1.0]
+FLAG_ANGULAR_CEN_X           -1.0         # x center coordinate for OPT__FLAG_ANGULAR (< 0=auto -> box center) [-1.0]
+FLAG_ANGULAR_CEN_Y           -1.0         # y center coordinate for OPT__FLAG_ANGULAR (< 0=auto -> box center) [-1.0]
+FLAG_ANGULAR_CEN_Z           -1.0         # z center coordinate for OPT__FLAG_ANGULAR (< 0=auto -> box center) [-1.0]
```
hyschive, line 172:
```suggestion
FLAG_ANGULAR_CEN_X           -1.0         # x center coordinate for OPT__FLAG_ANGULAR (<0=auto -> box center) [-1.0]
FLAG_ANGULAR_CEN_Y           -1.0         # y center coordinate for OPT__FLAG_ANGULAR (<0=auto -> box center) [-1.0]
FLAG_ANGULAR_CEN_Z           -1.0         # z center coordinate for OPT__FLAG_ANGULAR (<0=auto -> box center) [-1.0]
```

example/test_problem/Template/Input__Parameter
```diff
@@ -167,13 +167,13 @@ OPT__FLAG_USER                0           # flag: user-defined (Input__Flag_User
...
 OPT__FLAG_RADIAL              0           # flag: radial resolution (Input__Flag_RadialResolution) [0]
-RADIAL_CEN_X                 -1.0         # x coordinate of radial resolution center (if one of RADIAL_CEN_X/Y/Z < 0 -> box center) [-1.0]
-RADIAL_CEN_Y                 -1.0         # y coordinate of radial resolution center (if one of RADIAL_CEN_X/Y/Z < 0 -> box center) [-1.0]
-RADIAL_CEN_Z                 -1.0         # z coordinate of radial resolution center (if one of RADIAL_CEN_X/Y/Z < 0 -> box center) [-1.0]
+FLAG_RADIAL_CEN_X            -1.0         # x center coordinate for OPT__FLAG_RADIAL (< 0=auto -> box center) [-1.0]
+FLAG_RADIAL_CEN_Y            -1.0         # y center coordinate for OPT__FLAG_RADIAL (< 0=auto -> box center) [-1.0]
+FLAG_RADIAL_CEN_Z            -1.0         # z center coordinate for OPT__FLAG_RADIAL (< 0=auto -> box center) [-1.0]
```
hyschive, line 176:
```suggestion
FLAG_RADIAL_CEN_X            -1.0         # x center coordinate for OPT__FLAG_RADIAL (<0=auto -> box center) [-1.0]
FLAG_RADIAL_CEN_Y            -1.0         # y center coordinate for OPT__FLAG_RADIAL (<0=auto -> box center) [-1.0]
FLAG_RADIAL_CEN_Z            -1.0         # z center coordinate for OPT__FLAG_RADIAL (<0=auto -> box center) [-1.0]
```

hyschive (commented):
I've added some final minor comments before merging

ChunYen-Chen:
@hyschive @hfhsieh Thanks for the review. I have addressed all comments.

hyschive (approved):

PR #313, opened 2024-05-22
Fix inconsistent CoM position

### LINKED ISSUE #312: Inconsistent CoM from restart file
The CoM position is not reproducible from the restart file since the OpenMP loop used `schedule( runtime )` instead of `schedule( static )` to reduce the weighted value.

ChunYen-Chen:
Resolve #312. Thanks @hsinhaoHHuang for the useful discussion.

### DIFF
--- a/src/Auxiliary/Aux_FindWeightedAverageCenter.cpp
+++ b/src/Auxiliary/Aux_FindWeightedAverageCenter.cpp
@@ -209,6 +209,7 @@ void Aux_FindWeightedAverageCenter( double WeightedAverageCenter[], const double
                                   MinDens_No, MinPres_No, MinTemp_No, MinEntr_No, DE_Consistency_No );
             }
 
+//          use the "static" schedule for reproducibility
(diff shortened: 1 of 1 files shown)


src/Auxiliary/Aux_FindWeightedAverageCenter.cpp
```diff
@@ -209,7 +209,7 @@ void Aux_FindWeightedAverageCenter( double WeightedAverageCenter[], const double
                                   MinDens_No, MinPres_No, MinTemp_No, MinEntr_No, DE_Consistency_No );
             }
 
-#           pragma omp parallel for reduction ( +:W_ThisRank, WX_ThisRank, WY_ThisRank, WZ_ThisRank ) schedule( runtime )
+#           pragma omp parallel for reduction ( +:W_ThisRank, WX_ThisRank, WY_ThisRank, WZ_ThisRank ) schedule( static )
```
hyschive, line 212: Add the comment "//       use the "static" schedule for reproducibility"

ChunYen-Chen: Done!

hyschive (approved):

PR #314, opened 2024-05-22
Store conserved variable values to HDF5

### LINKED ISSUE #305: Store the reference values in Aux_Check_Conservation.cpp
### Issue
- After restarting a simulation, errors recorded by `OPT__CK_CONSERVATION` will be wrong due to incorrect reference values `*_Ref` in `Aux_Check_Conservation.cpp`.

### Possible solutions
- (Solution 1) Record these reference values in the HDF5 snapshots `Data_XXXXXX` (in `KeyInfo_t`)
  - Pros: maintain double precision
  - Cons: not support `OPT__OUTPUT_TOTAL = 2` (which is fine since it's deprecated anyway)
- (Solution 2) Use the data at step 0 in `Record__Conservation` to compute the errors
  - Pros: doesn't rely on `OPT__OUTPUT_TOTAL`
  - Cons: lose double precision unless we output more digits
- (Solution 3) Output a separate small HDF5 file (by utilizing #304) to store these reference values when calling `Aux_Check_Conservation()` for the first time.

I prefer Solution 1 since it's more consistent with the current restarting mechanism of GAMER.

ChunYen-Chen:
Use the solution 1 mentioned in #305 to resolve #305. Now, `*_Ref` is assigned only at the start of the simulation, otherwise, it is loaded from HDF5.


hsinhaoHHuang (approved):
I have run some tests using `BlastWave`, `ClusterMerger`, `CR_Diffusion`, `JetICMWall`, `MHD_OrszagTangVortex` and `Plummer` to check values in HDF5 data and check the `Record__Conservation` can be reproduced after restart.
There is only one small issue:
- For the `BlastWave` test problem (serial by default) with `--bitwise_reproducibility=true` and `OPT__CK_CONSERVATION` turned on. The `Record__Conservation` of the restarted run is different (order of machine precision) from the `Record__Conservation`  of the fresh run.
    - However, if the tests are run with MPI (`--mpi=true`), then the restarted `Record__Conservation` is the same as the fresh run.
    - This issue also happens in the main branch: The recorded total momentum is different after the restart.

And I have one minor suggestion:
- Since the variables `XXX_Ref` and `XXX_H5Ref` have become global and are not inside `Aux_Check_Conservation()`, how about adjusting the variable names or adding some comments to describe what kind of references they are to avoid confusion?

ChunYen-Chen:
Add a simple check of storing the reference value at `OutputDumpData_HDF5.cpp`.

ChunYen-Chen:
@hsinhaoHHuang The file `Aux_Check_Conservation.cpp` has been updated to the latest version. Please review it, thanks!

ChunYen-Chen:
> Add a simple check of storing the reference value at `OutputDumpData_HDF5.cpp`.

Done!

ChunYen-Chen:
> Add a simple check of storing the reference value at `OutputDumpData_HDF5.cpp`.

Done!

ChunYen-Chen:
@hsinhaoHHuang I have added the force record reference value code.

src/Output/Output_DumpData_Total_HDF5.cpp
```diff
@@ -1588,6 +1589,96 @@ void FillIn_KeyInfo( KeyInfo_t &KeyInfo, const int NFieldStored )
...
+   const int    NVar_NoPassive    = 11;
+   const int    idx_etot          = NVar_NoPassive - 1;
+#  endif
+   const bool GetPassiveSum = ( PassiveNorm_NVar > 0 );
+   const int  NVar          = NVar_NoPassive + NCOMP_PASSIVE + ( (GetPassiveSum)?1:0 );
+
+   if ( !ConservedRefLoaded )   Aux_Error( ERROR_INFO, "Conserved variables reference are not assigned yet!!\n" );
+   KeyInfo.Time_H5Ref        = Time_Ref;
+#  if   ( MODEL == HYDRO )
+   KeyInfo.Mass_Gas_H5Ref    = Fluid_Ref[0];
```
hsinhaoHHuang, line 1610: I still think we should include keywords related to "conservation" or "total" in these variable names. Otherwise, it is hard to understand the type of reference when someone looks at the hdf5 data.

hsinhaoHHuang (changes requested):
After you latest changes, it (`Plummer` test problem) will show this error if `OPT__CK_CONSERVATION` is off.
```
[eureka04:64557] *** An error occurred in MPI_Allreduce
[eureka04:64557] *** reported by process [140131862052865,1719644775776256]
[eureka04:64557] *** on communicator MPI_COMM_WORLD
[eureka04:64557] *** MPI_ERR_TRUNCATE: message truncated
[eureka04:64557] *** MPI_ERRORS_ARE_FATAL (processes in this communicator will now abort,
[eureka04:64557] ***    and potentially your MPI job)
```

ChunYen-Chen:
@hsinhaoHHuang Thanks for reviewing. I have fixed the bug caused by `ConservedRefLoaded` not being updated at non-master rank.

src/Auxiliary/Aux_Check.cpp
```diff
@@ -11,45 +11,46 @@ void Aux_Check( )
...
-      for (int lv=0; lv<NLEVEL-1; lv++)   Aux_Check_Refinement( lv, "DIAGNOSIS" );
+      for (int lv=0; lv<NLEVEL-1; lv++)                  Aux_Check_Refinement( lv, "DIAGNOSIS" );
 
    if ( OPT__CK_PROPER_NESTING )
-      for (int lv=1; lv<NLEVEL; lv++)     Aux_Check_ProperNesting( lv, "DIAGNOSIS" );
+      for (int lv=1; lv<NLEVEL; lv++)                    Aux_Check_ProperNesting( lv, "DIAGNOSIS" );
 
-   if ( OPT__CK_CONSERVATION )            Aux_Check_Conservation( "DIAGNOSIS" );
+   printf("Rank: %d, ConservedRefLoaded: %d\n", MPI_Rank, ConservedRefLoaded);
+   if ( OPT__CK_CONSERVATION || ! ConservedRefLoaded )   Aux_Check_Conservation( "DIAGNOSIS" );
```
hsinhaoHHuang, line 20: Although not always, it seems that the syntax with two spaces before and after `||` in the if statement is more common in GAMER.

hsinhaoHHuang (changes requested):
Thanks for the update!
I have finished further tests on restarting simulations from runs with `OPT__CK_CONSERVATION` on or off initially. The output `Record__Conservation` and those stored reference values in HDF5 data are all the same as the fresh runs.

I only have one final suggestion:
- Since `ANGMOM_ORIGIN_*` will always be used to calculate the conservation reference, how about also always recording them in `Record__Note` at [src/Auxiliary/Aux_TakeNote.cpp Line.1545](https://github.com/ChunYen-Chen/gamer-fork/blob/8fdd009b2d49522109f6eac64827695f7aaa39bd/src/Auxiliary/Aux_TakeNote.cpp#L1545)? It is possible that `OPT__CK_CONSERVATION` is off at first and is turned on during the restart, and user may want to look up what origin (which could possibly be reset during restart) is used to calculate the angular momenta reference.

ChunYen-Chen:
@hsinhaoHHuang Thanks for the review again! I have updated the code according to your suggestions.

hsinhaoHHuang (approved):

src/Auxiliary/Aux_Check_Conservation.cpp
hsinhaoHHuang, line 1: I just noticed we should also update the second point in the Note (Line 21 - Line 24.).

ChunYen-Chen: Updated.

hyschive:
@ChunYen-Chen Please resolve the conflicts.

hyschive (commented):
- [ ] The following hard-coded indices have been increasingly annoying as they ppear in multiple places. 
```
// ugly index from Aux_Check_Conservation.cpp
#  if   ( MODEL == HYDRO )
#  ifdef MHD
   const int    NVar_NoPassive    = 12;
#  else
   const int    NVar_NoPassive    = 11;
#  endif
   const int    idx_etot          = NVar_NoPassive - 1;
#  elif ( MODEL == ELBDM )
   const int    NVar_NoPassive    = 5;
   const int    idx_etot          = NVar_NoPassive - 1;
#  endif
   const bool GetPassiveSum = ( PassiveNorm_NVar > 0 );
   const int  NVar          = NVar_NoPassive + NCOMP_PASSIVE + ( (GetPassiveSum)?1:0 );
```
One possible solution is to introduce a single global variable `ConRef[NCONREF_MAX+NCOMP_PASSIVE]` to store all the reference conserved values, including time.
- `NCONREF_MAX` is a symbolic constant in `Macro.h`, set to a fixed value (e.g., `30`)
- Initialize this array as -1
- In `Output_DumpData_Total_HDF5.cpp` and `Init_ByRestart_HDF5.cpp`, simply output and load all `NCONREF_MAX+NCOMP_PASSIVE` variables
- Keep all the hard-coded indices, such as `NVar_NoPassive, idx_etot, NVar`, within `Aux_Check_Conservation.cpp` to minimize changes elsewhere

src/Init/Init_ResetParameter.cpp
```diff
@@ -811,28 +811,25 @@ void Init_ResetParameter()
 
 
 // set default value for the origin of angular momentum
-   if ( OPT__CK_CONSERVATION )
```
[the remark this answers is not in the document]
ChunYen-Chen: Since we always record `AngMom*` in this PR, the origin of the angular momentum must be declared.

src/Main/Main.cpp
```diff
@@ -93,6 +93,10 @@ OptLohnerForm_t      OPT__FLAG_LOHNER_FORM;
 OptCorrAfterSync_t   OPT__CORR_AFTER_ALL_SYNC;
 OptTimeStepLevel_t   OPT__DT_LEVEL;
 
+bool                 ConservedRefInitialized = false;
+double               ConservedRef[1+NCONREF_MAX+NCOMP_PASSIVE+1];
```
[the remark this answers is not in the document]
hyschive: OK. Add a comment on this extra `1` when defining `ConRef` in `main.cpp`.

src/Auxiliary/Aux_Check_Conservation.cpp
```diff
@@ -401,131 +414,141 @@ void Aux_Check_Conservation( const char *comment )
...
-      static double MomX_All_Ref, MomY_All_Ref, MomZ_All_Ref;
-      static double AngMomX_All_Ref, AngMomY_All_Ref, AngMomZ_All_Ref, Etot_All_Ref;
-#     endif
-#     endif // #ifdef PARTICLE
-      double AbsErr[NVar], RelErr[NVar];
-
-      if ( FirstTime )
+//    record the reference values if not initialized, e.g. first time, not from restart, or HDF5 version < 2502
+      if ( ! ConservedRefInitialized )
       {
```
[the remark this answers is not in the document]
hyschive: Quantities like total momentum can be exactly zero in many tests. Alternatively, how about initializing it as `NULL_REAL`? The exact value doesn't really matter since the main purpose is to let data stored in `Data_*` be deterministic.

src/Auxiliary/Aux_Check_Conservation.cpp
```diff
@@ -401,131 +414,136 @@ void Aux_Check_Conservation( const char *comment )
...
+//    record the reference values if not initialized, e.g., first time or restart from an HDF5 snapshot with version < 2502
+      if ( ! ConRefInitialized )
       {
-//       record the reference values
-         Time_Ref = Time[0];
-         for (int v=0; v<NVar; v++)    Fluid_Ref[v]   = Fluid_AllRank[v];
-         for (int d=0; d<3; d++)       CoM_Gas_Ref[d] = CoM_Gas[d];
+         ConRef[0] = Time[0];
+         for (int v=0; v<NVar_Flu; v++)   ConRef[idx_offset_flu    +v] = Fluid_AllRank[v];
+         for (int d=0; d<3; d++)          ConRef[idx_offset_flu_com+d] = CoM_Flu[d];
```
hyschive, line 502: Can we add several sanity checks below to detect overflow? They could be similar to `Aux_Error( ERROR_INFO, "exceed NFIELD_STORED_MAX (%d) !!\n", NFIELD_STORED_MAX );` in `Output_DumpData_Total_HDF5.cpp`

src/Auxiliary/Aux_Check_Conservation.cpp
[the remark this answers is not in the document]
ChunYen-Chen: Got it.

src/Main/Main.cpp
```diff
@@ -93,9 +93,8 @@ OptLohnerForm_t      OPT__FLAG_LOHNER_FORM;
 OptCorrAfterSync_t   OPT__CORR_AFTER_ALL_SYNC;
 OptTimeStepLevel_t   OPT__DT_LEVEL;
 
-bool                 ConservedRefInitialized = false;
-double               ConservedRef[1+NCONREF_MAX+NCOMP_PASSIVE+1];
-
+bool                 ConRefInitialized = false;
+double               ConRef[1+NCONREF_MAX+NCOMP_PASSIVE]; // time + conserved variables + passive scalars
```
hyschive, line 97: I prefer including `NCOMP_PASSIVE` within `NCONREF_MAX` (i.e., `ConRef[1+NCONREF_MAX]`) for simplicity (or even just `ConRef[NCONREF_MAX]`. This will also make the newly added sanity check in `Aux_Check_Conservation()` more intuitive. Let's discuss it in person.

src/Auxiliary/Aux_Check_Conservation.cpp
```diff
@@ -103,7 +103,10 @@ void Aux_Check_Conservation( const char *comment )
    const bool GetPassiveSum      = ( PassiveNorm_NVar > 0 );
    const int  NVar_Flu           = NVar_NoPassive + NCOMP_PASSIVE + ( (GetPassiveSum)?1:0 );
    const int  idx_offset_flu     = 1;
-   const int  idx_offset_flu_com = idx_offset_flu+NVar_Flu;
+   const int  idx_offset_flu_com = idx_offset_flu + NVar_Flu;
+
+   int NStoredConRef_noTime_noPassive = 0;
+   NStoredConRef_noTime_noPassive += NVar_Flu - NCOMP_PASSIVE + 3;
```
hyschive, line 109: Add the comment "// +3: center-of-mass position" here and for the following two updates of `NStoredConRef_noTime_noPassive` as well.

hyschive (commented):
Final comments added.

hyschive (approved):

PR #321, opened 2024-06-20
Improve the change parameter script

### LINKED ISSUE #311: Improve change_parameters.py
Add the flexibility to set certain parameter combinations but not to iterate over all permutations in `change_parameters.py`. For example, we may want to test `NX0_TOT_X/Y/Z=16/16/16` and `32/32/32` but not `16/32/16`.

ChunYen-Chen:
Resolve #311.

## Thanks
Thanks to @KunkelAlexander for providing this method.

## Add
Using `replace_parameter()` to change parameters by other parameters.

### DIFF
--- a/tool/simulation/change_parameters.py
+++ b/tool/simulation/change_parameters.py
@@ -135,24 +135,30 @@ def execution( **kwargs ):
     cwd               = os.getcwd()
     record_parameters = kwargs["record"]
 
-    # 1. Run gamer
+    # 1. Add parameter change by other parameters
+    record_parameters["NX0_TOT_Y"] = record_parameters["NX0_TOT_X"]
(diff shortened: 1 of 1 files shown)


vivi235711 (approved):

hyschive:
- IIUC, this PR doesn’t provide a general solution to address #311. What we need is support for a setup like `(a, b, c) = [(1, 2, 3), (4, 5, 6)]`, where `a, b, c` are arbitrary runtime options and the script should execute just two cases: `(a, b, c) = (1, 2, 3)` and `(a, b, c) = (4, 5, 6)`.
- Please clarify, presumably in the header, which test problem can be used to validate the default parameters defined in this script (perhaps `Plummer`?).

PR #333, opened 2024-07-06
Enable barotropic for isothermal EoS

hyschive:
Enable `--barotropic` automatically for `--eos=ISOTHERMAL` in `configure.py`

### DIFF
--- a/src/configure.py
+++ b/src/configure.py
@@ -447,7 +447,7 @@ def load_arguments():
                        )
 
     parser.add_argument( "--barotropic", type=str2bool, metavar="BOOLEAN", gamer_name="BAROTROPIC_EOS",
-                         default=False,
+                         default=None,
                          depend={"model":"HYDRO"},
                          constraint={ True:{"eos":["ISOTHERMAL", "TABULAR", "USER"]} },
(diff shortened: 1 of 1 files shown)


src/configure.py
```diff
@@ -744,6 +744,10 @@ def set_conditional_defaults( args ):
         if   args["cosmic_ray"]: args["eos"] = "COSMIC_RAY"
         elif args["srhd"]      : args["eos"] = "TAUBMATHEWS"
         else                   : args["eos"] = "GAMMA"
+
+    if args["barotropic"] is None:
+        if   args["eos"] == "ISOTHERMAL": args["barotropic"] = True
+        else                            : args["barotropic"] = False
```
ChunYen-Chen, line 750: How about `args["barotropic"] = (args["eos"] == "ISOTHERMAL")`?

ChunYen-Chen (approved):
Looks great! I only have one comment about reducing the code.

src/configure.py
hyschive: Updated

ChunYen-Chen (approved):

PR #357, opened 2024-09-29
Autocompletion of `configure.py`

ChunYen-Chen:
It is hard to memorize all the option names since `configure.py` has too many options. Now, you do not need to memorize the options anymore!
By simply pressing tab key, the available options will be shown and be autocompleted if there is only one available option.

### Other updates
* `-lh` shows the help message in alphabetical order.
* `Makefile.log` will not be generated with `-h`, `-lh`, and `--autocomplete_info` arguments.
* Add the verbose compilation mode argument `--verbose_make`
* The common prefix and suffix of the simulation options can be assigned in `parser.add_argument()`
* Separate long string


technic960183 (commented):
Thanks for the work on this PR.
After I try to use it, I found that it only support `--arg option`.
Is it possible to support `--arg=option`?
If `compgen` does not support it. Can we achieve this by modifing the  top level option when = is detected after a valid arg?

ChunYen-Chen:
@technic960183 Thanks for the comment and review. I have updated the code accordingly.

technic960183 (changes requested):
Thanks for the update, `--arg=option` works great. And I don't have any idea of a better implementation now.
Since our wiki use `python configure.py` as examples, I think support `python configure.py` is necessary.

tool/bash/config_autocomplete.sh
```diff
@@ -0,0 +1,59 @@
+#!/bin/bash
+
+__gamer_configure_autocomplete() {
+
+    local options option_array option_array2
```
technic960183, line 5: Add comments to explain `option_array` and `option_array2` for improving maintainability if you think that this is needed.

technic960183 (commented):
To avoid autocomplete being triggered on unrelated scripts named `configure.py`, I suggest verifying whether the configure.py being executed is the intended one.

I propose two possible solutions:

- **Use an Identifier for Validation**
Add or utilize an identifier to confirm the script. For example:

If `./configure.py -h` outputs `2023 Computational Astrophysics Lab, NTU. All rights reserved.` at the end, you can validate it by checking for `Computational Astrophysics Lab`.
Alternatively, introduce a new option such as `--version` and verify if the output matches a predefined pattern.

- **Verify Without Executing the Script**
This approach avoids calling `configure.py` altogether, which can be important if the other script has side effects (e.g., writing logs) when executed.

Use the `sha1sum` of `configure.py`, for instance:
```
026c866b02f873dad60290a43336ee555119a834  configure.py
```
Hard-code this hash into the autocomplete script and compare it. However, note that the hash would need to be updated every time `configure.py` is modified.
The above resolutions are suggestions. The choice of implementation is up to you and @hyschive to decide based on the project's priorities and constraints.

tool/bash/config_autocomplete.sh
technic960183, line 5: Add comments to explain `option_array` and `option_array2` for improving maintainability if you think that this is needed.

ChunYen-Chen: The variables are renamed.

ChunYen-Chen:
@technic960183 Thanks for the detailed comments. The `python configure.py` and `python3 configure.py` are now supported autocompletion.

My current strategy to verify the script is:
1. Check if this directory is a git directory.
2. Check if the very first commit hash is the GAMER first commit hash.
3. Check if the very first commit hash of `configure.py` is the `configure.py` for generating `Makefile`.

tool/bash/config_autocomplete.sh
```diff
@@ -0,0 +1,100 @@
...
+
+    # Not the GAEMR `configure.py`
+    if [ $? -ne 0 ]; then
+       COMPREPLY=( $( compgen -o default -- ${COMP_WORDS[COMP_CWORD]} ) )
+       # add a trailing space since `-o nospace`
+       for i in "${!COMPREPLY[@]}"; do
+           COMPREPLY[$i]="${COMPREPLY[$i]} "
+       done
+       return 0
+    fi
```
technic960183, line 47: 4 spaces for indentation.
```suggestion
        COMPREPLY=( $( compgen -o default -- ${COMP_WORDS[COMP_CWORD]} ) )
        # add a trailing space since `-o nospace`
        for i in "${!COMPREPLY[@]}"; do
            COMPREPLY[$i]="${COMPREPLY[$i]} "
        done
        return 0
```

tool/bash/config_autocomplete.sh
```diff
@@ -0,0 +1,100 @@
...
+
+    __gamer_check_gamer_info $configure_filename
+
+    # Not the GAEMR `configure.py`
+    if [ $? -ne 0 ]; then
+       COMPREPLY=( $( compgen -o default -- ${COMP_WORDS[COMP_CWORD]} ) )
+       # add a trailing space since `-o nospace`
+       for i in "${!COMPREPLY[@]}"; do
+           COMPREPLY[$i]="${COMPREPLY[$i]} "
+       done
```
technic960183, line 45: [BUG] This doesn't fully reporduce the default behavior

Steps for reproduce:
```bash
cd ~
git clone https://github.com/PrincetonUniversity/athena.git
cd athena
./configure.py <press tab>
```

Without our auto-complete:
```
CHANGELOG.md                 configure.py                 .git/                        inputs/                      setup.cfg
CITATION.cff                 CONTRIBUTING.md              .github/                     LICENSE                      src/
.codecov.yml                 CPPLINT.cfg                  .github_changelog_generator  Makefile.in                  tst/
CODE_OF_CONDUCT.md           doc/                         .gitignore                   README.md                    vis/
```

With our auto-complete:
```
CHANGELOG.md                  CONTRIBUTING.md               .github_changelog_generator   README.md 
CITATION.cff                  CPPLINT.cfg                   .gitignore                    setup.cfg 
.codecov.yml                  doc                           inputs                        src 
CODE_OF_CONDUCT.md            .git                          LICENSE                       tst 
configure.py                  .github                       Makefile.in                   vis
```

Another example:
```bash
./configure.py src/<press tab>
```

Without our auto-complete:
```
athena_arrays.hpp    cr/                  globals.cpp          mesh/                parameter_input.hpp  utils/
athena.hpp           crdiffusion/         globals.hpp          multigrid/           pgen/                
bvals/               defs.hpp.in          gravity/             nr_radiation/        reconstruct/         
chemistry/           eos/                 hydro/               orbital_advection/   scalars/             
chem_rad/            fft/                 inputs/              outputs/             task_list/           
coordinates/         field/               main.cpp             parameter_input.cpp  units/               
```

With our auto-complete:
```
src/athena_arrays.hpp     src/crdiffusion           src/gravity               src/orbital_advection     src/task_list 
src/athena.hpp            src/defs.hpp.in           src/hydro                 src/outputs               src/units 
src/bvals                 src/eos                   src/inputs                src/parameter_input.cpp   src/utils 
src/chemistry             src/fft                   src/main.cpp              src/parameter_input.hpp   
src/chem_rad              src/field                 src/mesh                  src/pgen                  
src/coordinates           src/globals.cpp           src/multigrid             src/reconstruct           
src/cr                    src/globals.hpp           src/nr_radiation          src/scalars               
```

tool/bash/config_autocomplete.sh
technic960183: Thanks. This comment could be resolved.

tool/bash/config_autocomplete.sh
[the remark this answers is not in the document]
technic960183: Reviewed. Works as expected inside GAMER. Thanks for the implementation.

technic960183 (changes requested):
Thanks for the contribution.
There are some bugs of side effects outside the GAMER directory related to falling back to the default behavior of auto-completition.
Checking the very first commit hash to verify the file is a brilliant idea!

ChunYen-Chen:
@technic960183 I can not reproduce the bugs you found. I guess these bugs are due to the `git` and `bash` versions. May I know your `git` and `bash` versions?

My environment:
- git version 1.8.3.1
- GNU bash, version 4.2.46(2)-release (x86_64-redhat-linux-gnu)

technic960183:
> @technic960183 I can not reproduce the bugs you found. I guess these bugs are due to the different version behavior. May I know your `git` and `bash` versions?
> 
> My environment:
> 
> * git version 1.8.3.1
> * GNU bash, version 4.2.46(2)-release (x86_64-redhat-linux-gnu)

By `git --version` and `bash --version` on our machine:
```
git version 2.34.1
GNU bash, version 5.1.16(1)-release (x86_64-pc-linux-gnu)
```

tool/bash/config_autocomplete.sh
ChunYen-Chen: @technic960183 I solved the inconsistent issue, but the side effect is that the command will not have the trailing blank.

Execution:
```bash
./configure.py CHANGELOG<tab>
```

Output:
```bash
./configure.py CHANGELOG.md
```

Original:
```bash
./configure.py CHANGELOG.md<space>
```

tool/bash/config_autocomplete.sh
```diff
@@ -0,0 +1,95 @@
...
+    local sub="${COMP_WORDS[COMP_CWORD-1]}"
+    local cur="${COMP_WORDS[COMP_CWORD]}"
+
+    if [[ "$first" == "python"* ]]; then
+        configure_filename=$second
+    else
+        configure_filename=$first
+    fi
+
+    __gamer_check_gamer_info $configure_filename
```
technic960183, line 37: Would you like to move the validation to the top, before assigning values to the local variables?

tool/bash/config_autocomplete.sh
technic960183: I have filed a PR (ChunYen-Chen/gamer-fork#8) to your branch for preventing the side effect. Please take a look.

tool/bash/config_autocomplete.sh
technic960183, line 1: [Discussion] Just a discussion:
I'm not sure the difference between `-o default` and `-o bashdefault`.
It seems that `bashdefault` is more general from the [doc](https://www.gnu.org/software/bash/manual/bash.html#Programmable-Completion-Builtins).
But when I try `-o bashdefault` it has no effect (not working).

I think that we don't need to consider the case of other program registering `python` or `./configure.py`. Maybe it is not worth to dig in.
But just for reference, maybe we can put a side note in our docs to say that we register 3 commands.
And also tell the user if they are using `python3.10` for example, they can register the command by themselves.
Maybe make a constant array on the top of the script for the registered commands?

technic960183 (commented):
Thanks for the bug fix.
I left some comments for improving efficiency and UX.

tool/bash/config_autocomplete.sh
ChunYen-Chen: Thanks for the solution!

tool/bash/config_autocomplete.sh
ChunYen-Chen: I also tried `-o bashdefault`, but it does not work for me.

> Maybe make a constant array on the top of the script for the registered commands?

Since the `complete` command only takes one line, I do not think it needs to make another constant array.

technic960183 (commented):
I found some possible edge cases and some consideration on the beginners.
Feel free to reject these suggestion.

doc/wiki/Installation.md
```diff
@@ -41,6 +41,45 @@ Furthermore, you can override the default setting by passing the [[--machine | I
...
+>    ```
+>    NOTE: There are three commands registered in the `config_autocomplete.sh`, `python` `python3` and `./configure.py`.
+>          If you would like to register your command, please add the command at the bottom of the script.
+>
+> 1. Reload `~/.bashrc` to enable the feature
+>    ```bash
+>    source ~/.bashrc
+>    ```
+>
+> Now, try to type `./configure.py` then press `<tab>` multiple times!
```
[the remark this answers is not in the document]
ChunYen-Chen: > After I build the wiki to preview. I feel that this tip section break the balance and the flow of reading a little bit.
As this is an optional feature, should we put it at the end and give it a section?

Agree.

tool/bash/config_autocomplete.sh
```diff
@@ -0,0 +1,97 @@
...
+        return 0
+    fi
+
+    local all_options sub_options all_option_array sub_option_array
+    local not_set=true
+    local subsub="${COMP_WORDS[COMP_CWORD-2]}"
+    local sub="${COMP_WORDS[COMP_CWORD-1]}"
+    local cur="${COMP_WORDS[COMP_CWORD]}"
+
+    all_options=$(${configure_command} --autocomplete_info=all)
```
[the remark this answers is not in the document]
ChunYen-Chen: I think it is because the shebang in `configure.py` is different from the `python` you used.
For example, the `python` in my command line is `/work1/chunyenc/Software/miniforge3/envs/all/bin/python`, but `./configure.py` used `/usr/bin/python3` to execute since I did not update the shebang to the `python` I used/loaded.

technic960183 (approved):
Tests done:

- [x] The installation process on Wiki.
- [x] Correct behavior of the autocompletion.
- [x] Does not affect other `python`, `python3` command.
- [x] Does not affect other `./configure.py` (in `athena` for example)

Although I approved this PR, please notice that a [bug](https://github.com/gamer-project/gamer/pull/357#discussion_r1925377920) can be triggered if there is an alias of `python`.

src/configure.py
```diff
@@ -196,6 +207,39 @@ def print_help( self, *args, **kwargs ):
...
+            trail_option = "=" if "type" in option else ""
+            if not any( target_option in [flag+trail_option, flag] for flag in option["flags"] ):
+                continue
+
+            # option with choices
+            if "choices" in option:
+                print( " ".join(option["choices"]) )
+                return
+
+            # help like option
```
hyschive, line 233:
```suggestion
            # help-like options
```

src/configure.py
```diff
@@ -196,6 +207,39 @@ def print_help( self, *args, **kwargs ):
...
+            config_files = list( map( lambda f: f.replace( ".config", "" ), config_files ) )
+            print( " ".join(config_files) )
+            return
+
+        for option in self.options:
+            trail_option = "=" if "type" in option else ""
+            if not any( target_option in [flag+trail_option, flag] for flag in option["flags"] ):
+                continue
+
+            # option with choices
```
hyschive, line 228:
```suggestion
            # options with choices
```

src/configure.py
```diff
@@ -574,7 +651,8 @@ def load_arguments( sys_setting : SystemSetting ):
     parser.add_argument( "--star_formation", type=str2bool, metavar="BOOLEAN", gamer_name="STAR_FORMATION",
                          default=False,
                          depend={"particle":True},
-                         help="Allow creating new particles after initialization. Must trun on <--store_pot_ghost> when <--store_par_acc> is adoped.\n"
+                         help="Allow creating new particles after initialization. "
+                              "Must trun on <--store_pot_ghost> when <--store_par_acc> is adoped.\n"
```
hyschive, line 655:
```suggestion
                              "Must turn on <--store_pot_ghost> when <--store_par_acc> is adoped.\n"
```

src/configure.py
```diff
@@ -1102,16 +1199,21 @@ def warning( paths, **kwargs ):
...
         LOGGER.warning("@@@%s@@@ is replaced to '' since the value is not given or the related option is disabled."%key)
 
-    # 5.3 Write
+    # 6.3 Write
     with open( GAMER_MAKE_OUT, "w") as make_out:
         make_out.write( command + makefile )
 
     LOGGER.info("========================================")
     LOGGER.info("%s is created."%GAMER_MAKE_OUT)
+    if args["verbose_make"]: LOGGER.info("%s is in verbose mode!"%GAMER_MAKE_OUT)
```
hyschive, line 1218:
```suggestion
    if args["verbose_make"]: LOGGER.info("%s is in verbose mode."%GAMER_MAKE_OUT)
```

hyschive (changes requested):
@ChunYen-Chen This is super useful! I've left some inline comments. Additionally, please address the following points before merging:
- [ ] Do not declare empty symbolic constants like `-DGPU_COMPUTE_CAPABILITY=` in the `Makefile` when disabling the corresponding options (e.g., `--gpu`).
- [ ] File an issue report for this [bug](https://calab-ntu.slack.com/archives/CMJ7B31AT/p1737819036800469?thread_ts=1736931684.077849&cid=CMJ7B31AT)
- [x] Several option choices have changed due to the newly added `prefix` feature (e.g., `--dual, --elbdm_scheme, --wave_scheme, --gramfe_scheme, --hybrid_scheme`). Please update both the Wiki page and `generate_make.sh` accordingly.

@technic960183 Thanks for the careful review!

hyschive (approved):

Issue #347, opened 2024-08-27
Increasing `GPU_NSTREAM` does not enhance performance


koarakawaii:
## Issues
* Increasing `GPU_NSTREAM` for `GAMER` running on `Spock` platform does not result in any performance gain (using `Perf_Overall` as metric), which is opposed to the observation in [`GAMER 2` paper](https://academic.oup.com/mnras/article/481/4/4815/5106358), Fig. 5.
* However, using multiple MPI ranks can harvest about `~1.9` times speed-up (see metric below). **Why MPI ranks can optimize GPU usage but streams cannot?**
![螢幕擷取畫面 2024-08-27 115932](https://github.com/user-attachments/assets/f1df06b4-8cef-43d8-8fae-9a943de10e6f)
>[!NOTE]
> [`K20X`](https://www.techpowerup.com/gpu-specs/tesla-k20x.c2315)'s architecture is `Kepler`
* Metric (run with `nsys profile` off):
    | `RankXThread ` |`GPU_NSTREAM`| Averaged `Perf_Overall` | Accumulated Time for `Flu_Adv` (s) | Accumulated Time Percentage for `Flu_Adv` |
    | :-----: | :-----: | :-----: | :-----: | :-----: |
    | 1X32 | 1 | 2.3668e+08 | 4100.43 | 87.787% |
    | 1X32 | 2 | 2.9351e+08 | 3221.56 | 85.395% |
    | 1X32 | 4 | 2.9881e+08 | 3165.45 | 85.466% |
    | 1X32 | 8 | 2.9150e+08 | 3238.99 | 85.236% |
    | 2X16 | 1 | 3.6865e+08 | 2495.44 | 83.168% |
    | 2X16 | 2 | 4.4064e+08 | 1999.24 | 79.495% |
    | 2X16 | 4 | 4.4752e+08 | 1934.79 | 78.141% |
    | 2X16 | 8 | 3.9993e+08 | 2231.59 | 79.581% |
    | **4X8**   | **1** | **4.5767e+08** | **1910.27** | **78.999%** |
    | 4X8   | 2 | 4.5726e+08 | 1919.73 | 79.298% |
    | 4X8   | 4 | 4.5165e+08 | 1920.74 | 78.380% |
    | 4X8   | 8 | 4.4038e+08 | 1932.00 | 76.873% |
    | 8X4   | 1 | 4.5478e+08 | 1913.43 | 78.635% |
[...]

PR #359, opened 2024-10-14
Set Default CUDA Stream Number as `4`

koarakawaii:
* Based on the [performance test](https://github.com/gamer-project/gamer/issues/347) and in consideration of striking the performance balance between `MPI RankPerNode=1` and `4`, we determined to set the default CUDA STREAM Number as `4` when `GPU_NSTREAM` is set to `-1` in `Input__Parameter`.
* When `MPI RankPerNode = 1`, warning message that suggests increasing `MPI RankPerNode` will be raised. See `src/MPI/Init_MPI.cpp`.


koarakawaii:
@ChunYen-Chen: thank you for all the advice~

hyschive:
@koarakawaii Could you update this PR to solve the conflict? Thanks!

hyschive (approved):
@koarakawaii Thanks for the contribution!

Issue #354, opened 2024-09-13
Grackle floating-point precision


hyschive:
- Add a compilation option `--double_grackle` to determine the floating-point precision of Grackle, similar to `--double_par` and `--double`. Note that `--double_grackle` and `--double` can be different.
- Adopt double precision by default as recommended in the [Grackle installation guide](https://grackle.readthedocs.io/en/latest/Installation.html).
- Declare `h_Che_Array` accordingly (or just use `gr_float` and then check the consistency between `gr_float` and `--double_grackle`)

ChunYen-Chen:
I think this issue is resolved.

PR #365, opened 2024-10-17
Grackle-related floating-point precision in GAMER

ChunYen-Chen:
## Issue
As pointed out by #354, the floating-point precision of Grackle (`gr_float` type) must be the same as GAMER floating-point precision (`real` type) now.

## Solution
This PR separates the Grackle-related variable type from `real` to `real_che` which is always the same as `gr_float`.

## How
Grackle determined the floating-point precision by `GRACKLE_FLOAT_4` and `GRACKLE_FLOAT_8` which are defined under `grackle_float.h`. This PR also uses `GRACKLE_FLOAT_4/8` to determine the type of `real_che` by including `grackle_float.h` in `include/Typedef.h`.

<details>
<summary> grackle_float.h (double precision)</summary>

```c++
#ifndef __GRACKLE_FLOAT_H__
#define __GRACKLE_FLOAT_H__
#define GRACKLE_FLOAT_8
#endif
```
</details>

## Performance on `spock`
Using `AGORA_IsolatedGalaxy` test problem. The table shows the accumulated time spent on each solver.
| Type(GAMER/Grackle) | Old dt | New dt | Old Flu | New Flu | Old Gra | New Gra | Old Che | New Che |
|---                  |---     |---     |---      |---      |---      |---      |---      |---      |
| single/single       | Fail   | Fail   | Fail    | Fail    | Fail    | Fail    | Fail    | Fail    |
| single/double       | X      | 312.01 | X       | 1016.88 | X       | 641.99  | X       | 1673.68 |
| double/single       | X      | Fail   | X       | Fail    | X       | Fail    | X       | Fail    |
| double/double       | 442.66 | 431.38 | 1944.36 | 1936.37 | 2635.52 | 2609.39 | 1467.52 | 1478.61 |
> [!NOTE]
> `AGORA_IsolatedGalaxy` fails when using the single precision Grackle.

As shown in the table, the new double-double precision performance is the same as the old run. 

### Result comparison
old_double_double
Star formation rate (old double/double)
new_double_double
Star formation rate (new double/double)
new_single_double
Star formation rate (new single/double)


include/Typedef.h
```diff
@@ -28,6 +28,19 @@ typedef double real_par;
...
+#include <grackle_float.h>
+#ifdef GRACKLE_FLOAT_8
+typedef double real_che;
+#else
+#ifdef GRACKLE_FLOAT_4
+typedef float  real_che;
+#else
+#error : GRACKLE_FLOAT_8 or GRACKLE_FLOAT_4 is not defined in grackle library!
+#endif
+#endif
```
hsinhaoHHuang, line 41: Would you think it would be simpler (and to avoid the nested if statements) to use:
```.h
#if   defined GRACKLE_FLOAT_8
...
#elif defined GRACKLE_FLOAT_4
...
#else
...
#endif
```

src/Grackle/Grackle_Init.cpp
```diff
@@ -28,12 +28,20 @@ void Grackle_Init()
 
 
 // check
-// floating-point type (don't know how to validate it yet...)
-   /*
-   if ( typeid(real) != typeid(gr_float) )
-      Aux_Error( ERROR_INFO, "inconsistent floating-point type (GAMER: %d, Grackle: %d) !!\n",
-                 sizeof(real), sizeof(gr_float) );
-                 */
+// floating-point type (You can not compile with different type)
```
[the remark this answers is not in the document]
ChunYen-Chen: No, this comment means the `real` and `gr_float` types, but this is out-dated. I will remove it.

src/Grackle/Grackle_Init.cpp
```diff
@@ -28,12 +28,20 @@ void Grackle_Init()
...
+/*
+   if ( typeid(real_che) != typeid(gr_float) )
+      Aux_Error( ERROR_INFO, "inconsistent floating-point type (GAMER(real_che): %d, Grackle: %d) !!\n",
+                 sizeof(real_che), sizeof(gr_float) );
+*/
```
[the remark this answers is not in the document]
ChunYen-Chen: Agree.

include/Typedef.h
ChunYen-Chen: Updated

ChunYen-Chen (commented):
@hsinhaoHHuang Thank you for the review. I have updated the codes accordingly.

hsinhaoHHuang (approved):

src/Grackle/Grackle_Init.cpp
```diff
@@ -28,12 +28,20 @@ void Grackle_Init()
...
-                 sizeof(real), sizeof(gr_float) );
-                 */
+// errors
+   if ( typeid(real_che) != typeid(gr_float) )
+      Aux_Error( ERROR_INFO, "inconsistent floating-point type (GAMER(real_che): %d, Grackle: %d) !!\n",
+                 sizeof(real_che), sizeof(gr_float) );
+
+// warnings
+   if ( MPI_Rank == 0  &&  sizeof(gr_float) != 8 )
+       Aux_Message( stderr, "WARNING : Grackle suggests floating-point number to be 8 bytes (GRACKLE_FLOAT_8) !!\n" );
```
hyschive, line 38:
```suggestion
       Aux_Message( stderr, "WARNING : Grackle recommends using double precision (GRACKLE_FLOAT_8) !!\n" );
```

hyschive (commented):
@ChunYen-Chen Look good! I've added some super minor inline comments. One more thing:
- [x] IIUC, the following lines in `GAMER.h` are no longer useful.
```
#ifdef FLOAT8
#  define CONFIG_BFLOAT_8
#else
#  define CONFIG_BFLOAT_4
#endif
```

hyschive (approved):

PR #374, opened 2024-11-20
[Bug fix] Not setting GPU related options when `--gpu=false`

ChunYen-Chen:


src/configure.py
```diff
@@ -753,14 +753,18 @@ def set_gpu( gpus, flags, args ):
...
-        return gpu_opts
+        raise ValueError("GPU_COMPUTE_CAPABILITY is not set in `../configs/%s.config`. See `../configs/template.config` for illustration."%args["machine"])
     compute_capability = int(compute_capability)
 
     if   compute_capability < 0:
-        compute_capability = get_gpu_compute_capability()
+        try:
+            compute_capability = get_gpu_compute_capability()
+        except:
+            raise ValueError("Fail to find GPU_COMPUTE_CAPABILITY automatically, please set it manually!")
```
hyschive, line 767: How about changing the error message as follows? Please test it.
```
"Fail to set GPU_COMPUTE_CAPABILITY automatically! Please set it manually in `../configs/%s.config`."%args["machine"]
```

hyschive (approved):

PR #385, opened 2024-12-10
[Bugfix] Missing preprocessor

ChunYen-Chen:
As title


hyschive (approved):
@ChunYen-Chen @hsinhaoHHuang Thanks for the bugfix!

PR #389, opened 2024-12-25
Reset END_T when using the dump table

hyschive:
For `END_T < 0.0`, reset the ending time as the time of the last dump. Otherwise, reset the ending time to the largest data dump time not greater than the input `END_T`.


vivi235711 (commented):
I've tested this, and it works as expected. There's only one point that needs confirmation.
By the way, I’ve also created another PR https://github.com/hyschive/gamer-fork/pull/144  related to the dump table. Please take a look if you have time.

vivi235711 (approved):

PR #393, opened 2024-12-31
Updates to support macOS

jzuhone:
Some support for macOS was originally provided by PR #2.

That was originally developed for Intel Macs. Since then, Macs with ARM architecture have arrived, and the hack used to get the CPU ID on Intel Macs does not work for ARM Macs. 

I spent quite a bit of time trying to find a cross-platform solution to get CPU and memory info but it turns out there isn't one and that not even the solution we came up with earlier is thought to be reliable. 

Since this is just for diagnostic information, I propose (and implement in this PR) that we disable CPU and memory reporting for macOS until we can find a solution. This will reduce lots of extraneous messages printed to stderr/stdout. 

Two other things:

* I found a solution to support both gcc and clang compilers. I have provided two different configs to this effect. 
* For whatever reason, the syntax `-rpath=/path/to/lib` is not supported on macOS and we must use `-rpath,/path/to/lib` instead. I believe that this is supposed to be cross-platform but need to check. 

Also, the C++11 `random` is the only one that is supported on macOS--should probably find a way to fix that switch in the case of this platform. Is there some way we can do that automatically?

TODO:

- [x] Test rpath logic on Linux
- [x] Find best way to enforce `RNG_CPP11` on macOS


hyschive:
@jzuhone Thank you for the PR!
@ChunYen-Chen Please help review and test it. Besides, do you have any suggestions for determining a suitable random number library automatically?

jzuhone:
@hyschive @ChunYen-Chen I verified that I could compile and run gamer using the `BlastWave` problem on Linux with this PR.

ChunYen-Chen:
@jzuhone @hyschive 

I think we can determine the `RANDOM_NUMBER` when generating the `Makefile` in `configure.py`. ([example](https://github.com/ChunYen-Chen/gamer-fork/blob/744af35300655861e8430aea5ceebde343c6c432/src/configure.py#L787-L789))
Also, we can raise an error when `#if ( defined ( __APPLE__ || __MACH__ )  &&  RANDOM_NUMBER != RNG_CPP11 )`.

jzuhone:
@ChunYen-Chen I've added some commits to this effect, thanks for the advice. 

I believe that this is ready for a final review.

src/Auxiliary/Aux_Check_MemFree.cpp
```diff
@@ -18,6 +18,11 @@
 void Aux_Check_MemFree( const double MinMemFree_Total, const char *comment )
 {
 
+#ifdef __APPLE__
+   // memory reporting is not currently supported on macOS
```
ChunYen-Chen, line 22: [Style]
```suggestion
// memory reporting is not currently supported on macOS
```

src/Main/Main.cpp
```diff
@@ -619,7 +619,13 @@ int main( int argc, char *argv[] )
    Output_DumpData( 0 );
 
    if ( OPT__PATCH_COUNT > 0 )            Aux_Record_PatchCount();
-   if ( OPT__RECORD_MEMORY )              Aux_GetMemInfo();
+   if ( OPT__RECORD_MEMORY )              
+#     ifdef __APPLE__
+      Aux_Message( stderr, "WARNING : memory reporting is not currently supported on macOS !!\n" );
+#     else
+      Aux_GetMemInfo();
+#     endif
```
ChunYen-Chen, line 627: How about adding the following lines in `src/Auxiliary/Aux_GetMemInfo.cpp` so that we do not have to update all the codes calling the function `Aux_GetMemInfo()`?
```c++
// memory reporting is not currently supported on macOS
#ifdef __APPLE__
return;
#endif
```

Also, please add the check of `OPT__RECORD_MEMORY` in `src/Auxiliary/Aux_Check_Parameter.cpp`.
```c++
#ifdef __APPLE__
if ( OPT__RECORD_MEMORY )
   Aux_Message( stderr, "WARNING : memory reporting is not currently supported on macOS !!\n" );
#endif
```

src/Main/Main.cpp
```diff
@@ -711,7 +717,10 @@ int main( int argc, char *argv[] )
       TIMING_FUNC(   Aux_Record_PatchCount(),         Timer_Main[4],   TIMER_ON   );
 
       if ( OPT__RECORD_MEMORY )
+#     ifndef __APPLE__
+      // memory reporting is not currently supported on macOS
       TIMING_FUNC(   Aux_GetMemInfo(),                Timer_Main[4],   TIMER_ON   );
+#     endif
```
ChunYen-Chen, line 723: Please check the comment on line 623.

include/RandomNumber.h
```diff
@@ -13,6 +13,10 @@
 #error : ERROR : unsupported RANDOM_NUMBER !!
 #endif
 
+#if ( ( defined (__APPLE__) || defined (__MACH__) )  &&  RANDOM_NUMBER != RNG_CPP11 )
+#error : ERROR : macOS requires RANDOM_NUMBER == RNG_CPP11 !!
+#endif
+
```
ChunYen-Chen, line 19: How about putting these codes in `src/Auxiliary/Aux_Check_Parameter.cpp`?

ChunYen-Chen (commented):
@jzuhone Thanks for the contribution. I have left some minor comments.

include/RandomNumber.h
jzuhone: I tried this and it didn't work because you get a compilation error related to random numbers even before you get to compiling `Aux_Check_Parameter.cpp`.

src/Main/Main.cpp
jzuhone: I've done this.

src/Main/Main.cpp
jzuhone: I've done this.

jzuhone:
@ChunYen-Chen I addressed your comments, thank you for them. There was one that I wasn't able to implement--see that one.

include/RandomNumber.h
ChunYen-Chen: I see. Thanks for the clarification.

ChunYen-Chen (approved):

hyschive (approved):
@jzuhone @ChunYen-Chen Looks good. Thanks for your contribution and review!

PR #399, opened 2025-01-04
[Bugfix] SwitchFinerLevelsToWaveScheme in Refine.cpp

hyschive:
This bugfix prevents `SwitchFinerLevelsToWaveScheme` from being reset to `false` in `Refine.cpp` in the serial mode. This issue happens when a fluid patch at the `ELBDM_FIRST_WAVE_LEVEL-1` level is flagged for refinement by criteria other than `OPT__FLAG_INTERFERENCE`. Note that the parallel routine `LB_Refine_GetNewRealPatchList.cpp` does not have this problem.


KunkelAlexander:
@hyschive Thank you for this bugfix and sorry for my late reply. That absolutely makes sense and I am sorry I had not thought of this before.

KunkelAlexander (approved):
Perfect, that is what this line was supposed to do. Thank you!

hyschive:
Thanks for the review!

PR #407, opened 2025-01-08
Add the runtime parameter `OPT__OUTPUT_TEXT_LENGTH_INT `

### LINKED ISSUE #401: Add a runtime parameter  `OPT__OUTPUT_TEXT_FORMAT_INT`
> Should we add a runtime parameter similar to `OPT__OUTPUT_TEXT_FORMAT_FLT` for integer variables? If you agree, I prefer to do that in a separate PR. What do you think?

_Originally posted by @hyschive in https://github.com/gamer-project/gamer/pull/348#discussion_r1898783391_

ChunYen-Chen:
Resolve #401.


hyschive (commented):
@ChunYen-Chen I have a potentially major concern about the incompatibility between different integer types (e.g., `int` and `long`) for variadic functions like `printf()`. For example, IIUC, the following code is invalid in general.
```
int a = 3;
printf( "%ld", a );
```
Given that `OPT__OUTPUT_TEXT_FORMAT_INT` can be set rather arbitrarily (e.g., `%12ld`, `%12d`, or `%12u`), it seems we cannot prevent such incompatibility issues from happening.

Possible solution:
- Allow users to specify only the length of the integer output (e.g., `OPT__OUTPUT_TEXT_LENGTH_INT = 12` )
- Always assume a long integer (`ld`) in the format specifier of `printf()`
- Explicitly cast integer arguments to `long` in function calls.

For example,
```
int temp = 3;
printf( "%*ld", OPT__OUTPUT_TEXT_LENGTH_INT, (long)temp );
```

What are your thoughts

include/HDF5_Typedef.h
```diff
@@ -831,6 +831,7 @@ struct InputPara_t
    int    Opt__Output_Step;
    double Opt__Output_Dt;
    char  *Opt__Output_Text_Format_Flt;
+   int    Opt__Output_Text_Length_Int;
```
hyschive, line 834: Add it to `Init_ByRestart_HDF5.cpp` as well for completeness.

example/test_problem/Template/Input__Parameter
```diff
@@ -380,6 +380,7 @@ SPEC_INT_GHOST_BOUNDARY       4           # ghost boundary size for spectral int
 OPT__OUTPUT_TOTAL             1           # output the simulation snapshot: (0=off, 1=HDF5, 2=C-binary) [1]
 OPT__OUTPUT_PART              0           # output a single line or slice: (0=off, 1=xy, 2=yz, 3=xz, 4=x, 5=y, 6=z, 7=diag) [0]
 OPT__OUTPUT_TEXT_FORMAT_FLT   %24.16e     # string format of output text files [%24.16e]
+OPT__OUTPUT_TEXT_LENGTH_INT   12          # string length of integer output text files [12]
```
hyschive, line 383: - For `OPT__OUTPUT_TEXT_FORMAT_FLT`, change the comment to "string format of floating-point variables in output text files.
- For `OPT__OUTPUT_TEXT_LENGTH_INT`, change the comment to "string length of integer variables in output text files"

hyschive (changes requested):
- [x] Apply `OPT__OUTPUT_TEXT_LENGTH_INT` to `Par_Output_TextFile.cpp`

ChunYen-Chen:
@hyschive Thanks for the review. I have updated the PR.

hyschive (approved):

PR #351, opened 2024-09-09
Add particle unique ID

ChunYen-Chen:
This is a follow-up PR of PR #348. It solves part of #79.

# Changes
* Wiki pages about particles are updated.

# New
* A new built-in particle attribute `PAR_PUID` to store the particle UID.
* A new section of Wiki pages about particle UID.
* A new script `track_particle.py` to track the particle by UID.

# Results
### Comparison of tracking particle by UID and index of the array with different MPI rank number
Here I show the results of three different UIDs (`300`, `5403`, and `31111`) with different MPI ranks (`1`, `2`, and `16`). As shown in the figures, the particles can be tracked well by the UID.
particle_00300
particle_05403
particle_31111\
The figure shows the position and the velocity in each direction of particles with specified UID. The `idx` means the index of specified particle UID in the array at step 0.

# Note
* `yt` module does not recognize the integer attribute type correctly.
* `gamer_compare_data` is not supported yet.
* `libyt` is also not supported yet.

Issue #135, opened 2022-08-07
Generalize fluid[]


hyschive:
- [ ] Support arbitrary cell-centered fields that are not passively advected.
  - [ ] Can be applied to `TEMP` and `DEDT` for the nuclear EoS (#51) and `POTE` for self-gravity
  - [ ] Can be used to combine the `HYDRO` and `ELBDM` models.

PR #420, opened 2025-01-26
Support using the Jeans length as the star formation threshold

hsinhaoHHuang:
## Goal
- To support using the local Jeans length $\lambda_J=\sqrt{\frac{\pi c_s^2}{G\rho}}$ as the star formation threshold

## Changes
- Generalize the original `SF_CreateStar_AGORA()` as `SF_CreateStar_GeneralGalaxy()` for general local star formation recipes.
  - Separate the part checking the star formation threshold to `SF_CreateStar_Check.cpp`
  - Separate the part deciding the star mass to `SF_CreateStar_GetStarMass.cpp`

- Add a new star formation scheme `SF_CREATE_STAR_SCHEME = 2` to use the Jeans length as the star formation threshold.

- Add a new runtime parameter `SF_CREATE_STAR_MAX_GAS_JEANSL`, where the star will only form when the local Jeans length is less than `dh`*`SF_CREATE_STAR_MAX_GAS_JEANSL`.

- Add a new test problem, `Hydro/StarFormationThreshold,` to test the implementation of star formation.
  - With varying temperature and density in the initial conditions.
  - Both particles and fluid are frozen and reset to initial conditions. Only stars keep forming.

## Verifications

- With the modifications, we can still reproduce the results of the original `Hydro/AGORA_IsolatedGalaxy` test problem (tested with `BITWISE_REPRODUCIBILITY` on).

- In the new `Hydro/StarFormationThreshold` test problem, the star forms in the expected region with the expected star mass and star formation rate.

  - Setups
    - BoxSize: 8 kpc
    - Grids: 128x128x32
    - `MAX_LEVEL                     0`
    - No initial particle
    - `OPT__FREEZE_PAR               1`
    - `OPT__FREEZE_FLUID             1`
    - `OPT__RESET_FLUID              1`
    - `SF_CREATE_STAR_MASS_EFF       1.0e-2`
    - `SF_CREATE_STAR_MIN_STAR_MASS  1.0e3 (Msun)`
    - `SF_CREATE_STAR_MAX_STAR_MFRAC 0.5`
 
  - Initial conditions
    - T-rho phase diagram = x-y spatial distribution
    image
 
  - `SF_CREATE_STAR_SCHEME = 1`, which uses density = `8.37e-24 g cm^-3` as the threshold
    - Star formation threshold
      image
      image
    - Star formation mass distribution
      image
    - Star formation rate
      image

  - `SF_CREATE_STAR_SCHEME = 2`, which uses Jeans length = `1 dh` as the threshold
    - Star formation threshold
      image
      image
    - Star formation mass distribution
      image
    - Star formation rate 
      image
       - The periodic pattern in the star formation rate seems related to the smaller time step close to the timing of Output_DumpData and the number of bins in the calculation. Due to (1) the large number of star-forming cells and (2) the total star masses dominated by the high-density cells, the variance due to the stochastic star formation is probably unobservable. For the cells (8~9th grids from the right) that will form a star with 50% probability, the theoretical sigma/mean = 1/sqrt(~32*96) < 2%.
    <details>
    <summary>Test with `OPT__RESET_FLUID = 0`, fluid was removed every step</summary>

     image

     image

     image

    </details>

---
### Comoving star formation tests (see README for the setup)
- `SF_CREATE_STAR_SCHEME = 1`
   - star formation rate
     image
   - accumulated stellar mass
     image

- `SF_CREATE_STAR_SCHEME = 2`
   - star formation rate
     image
   - accumulated stellar mass 
     image


ChunYen-Chen (changes requested):
@hsinhaoHHuang Thanks for the contribution. I have left some comments including some bugs. Also, please update your branch to solve the conflict.

hsinhaoHHuang (commented):
@ChunYen-Chen 
Thank you so much for your review. I have modified the code accordingly.

About adding the new runtime parameter in the `doc/wiki`, I plan to wait for #431 and update it (or I will update that PR if this PR is merged earlier), since all the star formation runtime parameters are completed there.

ChunYen-Chen (approved):

include/Typedef.h
```diff
@@ -93,6 +93,7 @@ const TestProbID_t
    TESTPROB_HYDRO_CR_SOUNDWAVE                 =   20,
    TESTPROB_HYDRO_CR_SHOCKTUBE                 =   21,
    TESTPROB_HYDRO_CR_DIFFUSION                 =   23,
+   TESTPROB_HYDRO_STAR_FORMATION_THRESHOLD     =   24,
```
hyschive: Remember to update this number if #498 is merged first.

src/Makefile_base
```diff
@@ -333,7 +333,7 @@ endif # SUPPORT_GRACKLE
 # star formation source files (included only if "STAR_FORMATION" is turned on)
 # ------------------------------------------------------------------------------------
 ifeq "$(filter -DSTAR_FORMATION, $(SIMU_OPTION))" "-DSTAR_FORMATION"
-CPU_FILE    += SF_CreateStar.cpp  SF_CreateStar_AGORA.cpp
+CPU_FILE    += SF_CreateStar.cpp  SF_CreateStar_GeneralGalaxy.cpp SF_CreateStar_Check.cpp SF_CreateStar_GetStarMass.cpp
```
hyschive:
```suggestion
CPU_FILE    += SF_CreateStar.cpp  SF_CreateStar_GeneralGalaxy.cpp  SF_CreateStar_Check.cpp  SF_CreateStar_GetStarMass.cpp
```

src/Auxiliary/Aux_Check_Parameter.cpp
```diff
@@ -1853,9 +1853,18 @@ void Aux_Check_Parameter()
...
+      Aux_Message( stderr, "WARNING : SF_CREATE_STAR_SCHEME == %d is incompatible with JEANS_MIN_PRES !!\n",
+                   SF_CREATE_STAR_SCHEME_DWARFGALAXY );
+      Aux_Message( stderr, "          --> SF_CREATE_STAR_MAX_GAS_JEANSL = %14.8e, but JEANS_MIN_PRES_NCELL = %d and JEANS_MIN_PRES_LEVEL = %d\n",
+                   SF_CREATE_STAR_MAX_GAS_JEANSL, JEANS_MIN_PRES_NCELL, JEANS_MIN_PRES_LEVEL );
+   }
+
+   if ( ( SF_CREATE_STAR_SCHEME == SF_CREATE_STAR_SCHEME_AGORA  ||  SF_CREATE_STAR_SCHEME == SF_CREATE_STAR_SCHEME_DWARFGALAXY )
+        &&  !SF_CREATE_STAR_DET_RANDOM )
+   {
+      Aux_Message( stderr, "WARNING : SF_CREATE_STAR_SCHEME == 1 or 2 will break bitwise reproducibility due to the\n" );
```
hyschive:
```suggestion
      Aux_Message( stderr, "WARNING : SF_CREATE_STAR_SCHEME == %d or %d will break bitwise reproducibility due to the\n", SF_CREATE_STAR_SCHEME_AGORA, SF_CREATE_STAR_SCHEME_DWARFGALAXY );
```

src/StarFormation/SF_CreateStar_GeneralGalaxy.cpp
```diff
@@ -6,37 +6,26 @@
 
 
 //-------------------------------------------------------------------------------------------------------
-// Function    :  SF_CreateStar_AGORA
-// Description :  Create new star particles stochastically using the presription suggested by the AGORA project
+// Function    :  SF_CreateStar_GeneralGalaxy
+// Description :  Create new star particles in the general methods where ther star formation is independent
```
hyschive: Besides the following minor modifications, what do you mean by "independent" here?
```suggestion
// Description :  Create new star particles using general methods where star formation is independent
```

src/StarFormation/SF_CreateStar_GeneralGalaxy.cpp
```diff
@@ -6,37 +6,26 @@
...
-//                MaxStarMFrac : Maximum gas mass fraction allowed to convert to stars          (--> "SF_CREATE_STAR_MAX_STAR_MFRAC")
-//                DetRandom    : Make random numbers determinisitic                             (--> "SF_CREATE_STAR_DET_RANDOM"    )
-//                UseMetal     : Store the metal mass fraction in star particles
+// Parameter   :  lv             : Target refinement level
+//                TimeNew        : Current physical time (after advancing solution by dt)
+//                dt             : Time interval to advance solution
+//                                 --> Currently this function does not distinguish dt and the physical time interval (dTime)
+//                                 --> Does NOT support COMOVING yet
+//                RNG            : Random number generator
+//                UseMetal       : Store the metal mass fraction in star particles
```
hyschive:
```suggestion
//                UseMetal : Store the metal mass fraction in star particles
```

src/StarFormation/SF_CreateStar_GeneralGalaxy.cpp
```diff
@@ -6,37 +6,26 @@
...
-//                MinStarMass  : Minimum star particle mass for the stochastical star formation (--> "SF_CREATE_STAR_MIN_STAR_MASS" )
-//                MaxStarMFrac : Maximum gas mass fraction allowed to convert to stars          (--> "SF_CREATE_STAR_MAX_STAR_MFRAC")
-//                DetRandom    : Make random numbers determinisitic                             (--> "SF_CREATE_STAR_DET_RANDOM"    )
-//                UseMetal     : Store the metal mass fraction in star particles
+// Parameter   :  lv             : Target refinement level
+//                TimeNew        : Current physical time (after advancing solution by dt)
+//                dt             : Time interval to advance solution
+//                                 --> Currently this function does not distinguish dt and the physical time interval (dTime)
+//                                 --> Does NOT support COMOVING yet
+//                RNG            : Random number generator
```
hyschive:
```suggestion
// Parameter   :  lv       : Target refinement level
//                TimeNew  : Current physical time (after advancing solution by dt)
//                dt       : Time interval to advance solution
//                           --> Currently this function does not distinguish dt and the physical time interval (dTime)
//                           --> Does NOT support COMOVING yet
//                RNG      : Random number generator
```

src/StarFormation/SF_CreateStar_GeneralGalaxy.cpp
hyschive, line 300: This additional comment is for future implementation, especially when addressing #135.
```suggestion
//###REVISE: skip non-density fields (e.g., temperature, cooling time, potential)
         for (int v=0; v<NCOMP_TOTAL; v++)   fluid[v][k][j][i] *= GasMFracLeft;
```

src/StarFormation/SF_CreateStar_GeneralGalaxy.cpp
```diff
@@ -294,6 +356,11 @@ void SF_CreateStar_AGORA( const int lv, const real TimeNew, const real dt, Rando
    } // for (int PID=0; PID<amr->NPatchComma[lv][1]; PID++)
 
 // free memory
+#  ifdef MHD
+   if ( NeedPres )   delete [] MagCC;
```
hyschive:
```suggestion
   delete [] MagCC;
```

src/StarFormation/SF_CreateStar_GeneralGalaxy.cpp
```diff
@@ -294,6 +356,11 @@ void SF_CreateStar_AGORA( const int lv, const real TimeNew, const real dt, Rando
...
+#  endif
+   if ( NeedPres )   delete [] Pres;
+   if ( NeedCs2  )   delete [] Cs2;
```
hyschive:
```suggestion
   delete [] Pres;
   delete [] Cs2;
```

src/StarFormation/SF_CreateStar_GeneralGalaxy.cpp
```diff
@@ -88,12 +74,27 @@ void SF_CreateStar_AGORA( const int lv, const real TimeNew, const real dt, Rando
 #  endif
 
    double x0, y0, z0, x, y, z;
-   real   GasDens, _GasDens, GasMass, _Time_FreeFall, StarMFrac, StarMass, GasMFracLeft;
+   real   GasDens, _GasDens, GasMass, StarMFrac, StarMass, GasMFracLeft;
    real   (*fluid)[PS1][PS1][PS1]      = NULL;
+   real   (*MagCC)[PS1][PS1][PS1]      = NULL;
```
hyschive:
```suggestion
#  ifdef MHD
   real   (*MagCC)[PS1][PS1][PS1]      = NULL;
#  endif
```

src/StarFormation/SF_CreateStar_GeneralGalaxy.cpp
```diff
@@ -88,12 +74,27 @@ void SF_CreateStar_AGORA( const int lv, const real TimeNew, const real dt, Rando
...
+   bool NeedCs2  = false;
+
+   if ( SF_CREATE_STAR_SCHEME == SF_CREATE_STAR_SCHEME_DWARFGALAXY )   NeedCs2 = true;
+   if ( NeedCs2 )   NeedPres = true;
+
+#  ifdef MHD
+   if ( NeedPres )   MagCC = new real [3][PS1][PS1][PS1];
+#  endif
+   if ( NeedPres )   Pres  = new real    [PS1][PS1][PS1];
+   if ( NeedCs2  )   Cs2   = new real    [PS1][PS1][PS1];
```
hyschive:
```suggestion
   if ( NeedPres )   MagCC = new real [NCOMP_MAG][PS1][PS1][PS1];
#  endif
   if ( NeedPres )   Pres  = new real            [PS1][PS1][PS1];
   if ( NeedCs2  )   Cs2   = new real            [PS1][PS1][PS1];
```

src/StarFormation/SF_CreateStar_GeneralGalaxy.cpp
```diff
@@ -129,6 +130,89 @@ void SF_CreateStar_AGORA( const int lv, const real TimeNew, const real dt, Rando
 #     ifdef STORE_POT_GHOST
       pot_ext = amr->patch[PotSg][lv][PID]->pot_ext;
 #     endif
+
+#     ifdef MHD
```
hyschive: How about changing to `#if ( defined MHD  &&  !defined DUAL_ENERGY )`?

src/StarFormation/SF_CreateStar_GeneralGalaxy.cpp
```diff
@@ -129,6 +130,89 @@ void SF_CreateStar_AGORA( const int lv, const real TimeNew, const real dt, Rando
...
+#           elif ( DUAL_ENERGY == DE_EINT )
+#           error : DE_EINT is NOT supported yet !!
+#           endif
+
+#           else // #ifdef DUAL_ENERGY
+
+#           ifdef MHD
+            const real Emag = (real)0.5*(  SQR( MagCC[MAGX][k][j][i] )
+                                         + SQR( MagCC[MAGY][k][j][i] )
+                                         + SQR( MagCC[MAGZ][k][j][i] )  );
```
hyschive: Since currently all we need here is the cell-centered magnetic energy, should we simply allocate `Emag[PS1][PS1][PS1]` and call `MHD_GetCellCenteredBEnergyInPatch()`? Or would you prefer to keep the current approach for generality?

src/StarFormation/SF_CreateStar_GeneralGalaxy.cpp
```diff
@@ -139,43 +223,21 @@ void SF_CreateStar_AGORA( const int lv, const real TimeNew, const real dt, Rando
...
+         if ( !SF_CreateStar_Check( lv, PID, i, j, k, dh, fluid, Pres, Cs2 ) )   continue;
 
+//       1-2. get the star mass
+         StarMass = SF_CreateStar_GetStarMass( GasDens, dv, dt, RNG, TID );
+         if ( StarMass <= 0.0 )   continue;
 
-//       1-5. check the maximum gas mass fraction allowed to convert to stars
-         StarMFrac = MIN( StarMFrac, MaxStarMFrac );
+//       check the maximum gas mass fraction allowed to convert to stars
+         StarMFrac = MIN( StarMass/GasMass, SF_CREATE_STAR_MAX_STAR_MFRAC );
```
hyschive:
```suggestion
         StarMFrac = FMIN( StarMass/GasMass, SF_CREATE_STAR_MAX_STAR_MFRAC );
```

src/StarFormation/SF_CreateStar_Check.cpp
```diff
@@ -0,0 +1,124 @@
...
+
+#if ( defined PARTICLE  &&  defined STAR_FORMATION  &&  MODEL == HYDRO )
+
+static bool SF_CreateStar_Check_GasDensity( const real GasDensity, const real Threshold );
+# ifdef GRAVITY
+static bool SF_CreateStar_Check_GasJeansLength( const real GasDensity, const real GasCs2, const real Threshold );
+# endif
+
+
+
```
hyschive:
```suggestion
#include "GAMER.h"

#if ( defined PARTICLE  &&  defined STAR_FORMATION  &&  MODEL == HYDRO )

static bool SF_CreateStar_Check_GasDensity( const real GasDensity, const real Threshold );
# ifdef GRAVITY
static bool SF_CreateStar_Check_GasJeansLength( const real GasDensity, const real GasCs2, const real Threshold );
# endif




```

src/StarFormation/SF_CreateStar_Check.cpp
```diff
@@ -0,0 +1,124 @@
...
+//-------------------------------------------------------------------------------------------------------
+// Function    :  SF_CreateStar_Check
+// Description :  Check if the target cell (i,j,k) satisfies the star formation criteria
+//
+// Note        :  1. Useless input arrays are set to NULL (e.g, Pot[] if GRAVITY is off)
```
hyschive:
```suggestion
// Note        :  1. Useless input arrays are set to NULL (e.g., Pot[] if GRAVITY is off)
```

src/StarFormation/SF_CreateStar_Check.cpp
```diff
@@ -0,0 +1,124 @@
...
+   bool AllowSF = true;
+
+   switch ( SF_CREATE_STAR_SCHEME )
+   {
+      case SF_CREATE_STAR_SCHEME_AGORA:
+//       create star particles only if the gas density exceeds the given threshold
+//       Ref: (1) Nathan Goldbaum, et al., 2015, ApJ, 814, 131 (arXiv: 1510.08458), sec. 2.4
+//            (2) Ji-hoon Kim, et al., 2016, ApJ, 833, 202 (arXiv: 1610.03066), sec. 3.2
+//       Currently this function does not check whether the cell mass exceeds the Jeans mass
+//       --> Ref: "jeanmass" in star_maker_ssn.F of Enzo
```
hyschive, line 46: Move it to the `Note` section of `SF_CreateStar_Check_GasDensity()`.

src/StarFormation/SF_CreateStar_Check.cpp
```diff
@@ -0,0 +1,124 @@
...
+// Parameter   :  lv            : Target refinement level
+//                PID           : Target patch ID
+//                i,j,k         : Indices of the target cell
+//                dh            : Cell size at the target level
+//                fluid         : Input fluid array (with NCOMP_TOTAL components)
+//                Pres          : Input pressure array
+//                Cs2           : Input squared sound speed array
+//
+// Return      :  "true"  if all  of the star formation criteria is satisfied
+//                "false" if none of the star formation criteria is satisfied
```
hyschive:
```suggestion
// Return      :  "true"  if the specified star formation criteria are satisfied
//                "false" otherwise
```

src/StarFormation/SF_CreateStar_Check.cpp
```diff
@@ -0,0 +1,124 @@
...
+   switch ( SF_CREATE_STAR_SCHEME )
+   {
+      case SF_CREATE_STAR_SCHEME_AGORA:
+//       create star particles only if the gas density exceeds the given threshold
+//       Ref: (1) Nathan Goldbaum, et al., 2015, ApJ, 814, 131 (arXiv: 1510.08458), sec. 2.4
+//            (2) Ji-hoon Kim, et al., 2016, ApJ, 833, 202 (arXiv: 1610.03066), sec. 3.2
+//       Currently this function does not check whether the cell mass exceeds the Jeans mass
+//       --> Ref: "jeanmass" in star_maker_ssn.F of Enzo
+         AllowSF &= SF_CreateStar_Check_GasDensity( fluid[DENS][k][j][i], SF_CREATE_STAR_MIN_GAS_DENS );
+         if ( !AllowSF )    return AllowSF;
```
hyschive: I'm a bit confused by the logic here. On one hand, you use `&=` and return immediately if `!AllowSF`, suggesting that stars form only when _all_ criteria are satisfied. On the other hands, different criteria are separated by `break`. 

I guess this is meant to accommodate future cases where a single `SF_CREATE_STAR_SCHEME` may invoke multiple `SF_CreateStar_Check_*()` routines. If so, could you clarify this in the comments?

src/StarFormation/SF_CreateStar_Check.cpp
```diff
@@ -0,0 +1,124 @@
...
+
+   if ( GasDensity >= Threshold )    AllowSF = true;
+
+   return AllowSF;
+
+} // FUNCTION : SF_CreateStar_Check_GasDensity
+
+
+
+#ifdef GRAVITY
```
hyschive: How about the following simplification?
- Enclose the line `real GasJeansL = SQRT( ( M_PI * GasCs2 ) / ( NEWTON_G * GasDensity ) );` with `#ifdef GRAVITY ... #endif`
- Raise an error at the beginning of this routine if `GRAVITY` is disabled.
- Remove all other `#ifdef GRAVITY ... #endif` guards around `SF_CreateStar_Check_GasJeansLength()` elsewhere.

By doing so, we don't need to add additional `#ifdef GRAVITY ... #endif` guards if this routine is invoked elsewhere.

src/StarFormation/SF_CreateStar_Check.cpp
```diff
@@ -0,0 +1,124 @@
...
+
+   return AllowSF;
+
+} // FUNCTION : SF_CreateStar_Check
+
+
+
+//-------------------------------------------------------------------------------------------------------
+// Function    :  SF_CreateStar_Check_GasDensity
+// Description :  Check if the gas desnity exceeds the given threshold
```
hyschive:
```suggestion
// Description :  Check if the gas density exceeds the given threshold
```

src/StarFormation/SF_CreateStar_Check.cpp
```diff
@@ -0,0 +1,124 @@
...
+//
+// Note        :  1.
+//
+// Parameter   :  GasDensity : Gas density
+//                Threshold  : Threshold for the star formation
+//
+// Return      :  "true"  if the gas density is larger           than the given threshold
+//                "false" if the gas density is equal or smaller than the given threshold
```
hyschive:
```suggestion
// Return      :  "true"  if the gas density is larger than or equal to the given threshold
//                "false" otherwise
```

src/StarFormation/SF_CreateStar_Check.cpp
```diff
@@ -0,0 +1,124 @@
...
+// Return      :  "true"  if the gas Jeans length is smaller         than the given threshold
+//                "false" if the gas Jeans length is equal or larger than the given threshold
+//-------------------------------------------------------------------------------------------------------
+bool SF_CreateStar_Check_GasJeansLength( const real GasDensity, const real GasCs2, const real Threshold )
+{
+   bool AllowSF = false;
+
+   real GasJeansL = SQRT( ( M_PI * GasCs2 ) / ( NEWTON_G * GasDensity ) );
+
+   if ( GasJeansL <= Threshold )    AllowSF = true;
```
hyschive: To improve performance
```suggestion
   const real GasJeansL2 = ( M_PI * GasCs2 ) / ( NEWTON_G * GasDensity );

   if ( GasJeansL2 <= SQR(Threshold) )    AllowSF = true;
```

src/StarFormation/SF_CreateStar_Check.cpp
```diff
@@ -0,0 +1,124 @@
...
+// Description :  Check if the gas Jeans length is below the given threshold
+//
+// Note        :  1. Gas Jeans length = \sqrt{ \frac{ \pi Cs^2 }{ G \rho } }
+//
+// Parameter   :  GasDensity : Gas density
+//                GasCs2     : Gas squared sound speed
+//                Threshold  : Threshold for the star formation
+//
+// Return      :  "true"  if the gas Jeans length is smaller         than the given threshold
+//                "false" if the gas Jeans length is equal or larger than the given threshold
```
hyschive:
```suggestion
// Return      :  "true"  if the gas Jeans length is smaller than or equal to the given threshold
//                "false" otherwise
```

src/StarFormation/SF_CreateStar_GetStarMass.cpp
```diff
@@ -0,0 +1,141 @@
...
+
+static real SF_CreateStar_GetStarMass_StochasticLoaclSchmidtLaw( const real GasDens, const real dv, const real dt, RandomNumber_t *RNG,
+                                                                 const real Efficiency, const real MinStarMass, const int TID );
+static real SF_CreateStar_GetStarMass_MaxStarM( const real GasDens, const real dv, const real MaxStarMFrac );
+
+
+
+//-------------------------------------------------------------------------------------------------------
+// Function    :  SF_CreateStar_GetStarMass
+// Description :  Get the mass of star according to the adopted models
```
hyschive:
```suggestion
// Description :  Determine the star particle mass according to the adopted star formation model
```

src/StarFormation/SF_CreateStar_GetStarMass.cpp
```diff
@@ -0,0 +1,141 @@
...
+//
+// Note        :  1. Return StarMass = -1.0 if the star is decided not to form (e.g., in the stochastic models )
```
hyschive:
```suggestion
// Note        :  1. Return StarMass = -1.0 if no star forms (e.g., in the stochastic model)
```

src/StarFormation/SF_CreateStar_GetStarMass.cpp
```diff
@@ -0,0 +1,141 @@
...
+//
+// Parameter   :  GasDens        : Gas density
+//                dv             : Cell volume
+//                dt             : Time interval to advance solution
+//                                 --> Currently this function does not distinguish dt and the physical time interval (dTime)
+//                                 --> Does NOT support COMOVING yet
+//                RNG            : Random number generator
+//                TID            : OpenMP thread ID
```
hyschive:
```suggestion
// Parameter   :  GasDens : Gas density
//                dv      : Cell volume
//                dt      : Time interval to advance solution
//                          --> Currently this function does not distinguish dt and the physical time interval (dTime)
//                          --> Does NOT support COMOVING yet
//                RNG    : Random number generator
//                TID     : OpenMP thread ID
```

src/StarFormation/SF_CreateStar_GetStarMass.cpp
```diff
@@ -0,0 +1,141 @@
...
+// Function    :  SF_CreateStar_GetStarMass
+// Description :  Get the mass of star according to the adopted models
+//
+// Note        :  1. Return StarMass = -1.0 if the star is decided not to form (e.g., in the stochastic models )
+//
+// Parameter   :  GasDens        : Gas density
+//                dv             : Cell volume
+//                dt             : Time interval to advance solution
+//                                 --> Currently this function does not distinguish dt and the physical time interval (dTime)
+//                                 --> Does NOT support COMOVING yet
```
hyschive: Update this PR to support comoving coordinates once #349 is merged.

src/StarFormation/SF_CreateStar_GetStarMass.cpp
```diff
@@ -0,0 +1,141 @@
...
+
+//-------------------------------------------------------------------------------------------------------
+// Function    :  SF_CreateStar_GetStarMass_MaxStarM
+// Description :  Maximum gas mass allowed to be converted to star
+//
+// Note        :  1.
+//
+// Parameter   :  GasDens        : Gas density
+//                dv             : Cell volume
+//                MaxStarMFrac   : Minimum star particle mass for the stochastical star formation (--> "SF_CREATE_STAR_MAX_STAR_MFRAC" )
```
hyschive:
```suggestion
//                MaxStarMFrac : Maximum gas mass fraction allowed to convert to stars (--> "SF_CREATE_STAR_MAX_STAR_MFRAC" )
```

src/StarFormation/SF_CreateStar_GetStarMass.cpp
```diff
@@ -0,0 +1,141 @@
...
+
+
+//-------------------------------------------------------------------------------------------------------
+// Function    :  SF_CreateStar_GetStarMass_MaxStarM
+// Description :  Maximum gas mass allowed to be converted to star
+//
+// Note        :  1.
+//
+// Parameter   :  GasDens        : Gas density
+//                dv             : Cell volume
```
hyschive:
```suggestion
// Parameter   :  GasDens      : Gas density
//                dv           : Cell volume
```

src/StarFormation/SF_CreateStar_GetStarMass.cpp
```diff
@@ -0,0 +1,141 @@
...
+
+
+   return StarMass;
+
+} // FUNCTION : SF_CreateStar_GetStarMass_StochasticLoaclSchmidtLaw
+
+
+
+//-------------------------------------------------------------------------------------------------------
+// Function    :  SF_CreateStar_GetStarMass_MaxStarM
```
hyschive, line 124: It appears that this routine is currently not used at all. Instead, you check `SF_CREATE_STAR_MAX_STAR_MFRAC` after calling `SF_CreateStar_GetStarMass()` in `SF_CreateStar_GeneralGalaxy()`.

I feel that the `SF_CREATE_STAR_MAX_STAR_MFRAC` criterion should be part of `SF_CreateStar_GetStarMass()` . If you agree, how about calling `SF_CreateStar_GetStarMass_MaxStarM()` at the end of `SF_CreateStar_GetStarMass()`, instead of inside `SF_CreateStar_GetStarMass_StochasticLoaclSchmidtLaw()`, so that it applies to other `SF_CreateStar_GetStarMass_*()` routines in the future? We then remove the corresponding code from `SF_CreateStar_GeneralGalaxy()`.

src/StarFormation/SF_CreateStar_GetStarMass.cpp
```diff
@@ -0,0 +1,141 @@
...
+
+   return StarMass;
+
+} // FUNCTION : SF_CreateStar_GetStarMass
+
+
+
+//-------------------------------------------------------------------------------------------------------
+// Function    :  SF_CreateStar_GetStarMass_StochasticLoaclSchmidtLaw
+// Description :  Decide the new star particles stochastically using the presription suggested by the AGORA project
```
hyschive:
```suggestion
// Description :  Determine the star particle mass stochastically following the AGORA project prescription
```

src/StarFormation/SF_CreateStar_GetStarMass.cpp
```diff
@@ -0,0 +1,141 @@
...
+//
+// Parameter   :  GasDens        : Gas density
+//                dv             : Cell volume
+//                dt             : Time interval to advance solution
+//                                 --> Currently this function does not distinguish dt and the physical time interval (dTime)
+//                                 --> Does NOT support COMOVING yet
+//                RNG            : Random number generator
+//                Efficiency     : Gas-to-star mass efficiency                                    (--> "SF_CREATE_STAR_MASS_EFF"      )
+//                MinStarMass    : Minimum star particle mass for the stochastical star formation (--> "SF_CREATE_STAR_MIN_STAR_MASS" )
+//                TID            : OpenMP thread ID
```
hyschive:
```suggestion
// Parameter   :  GasDens     : Gas density
//                dv          : Cell volume
//                dt          : Time interval to advance solution
//                              --> Currently this function does not distinguish dt and the physical time interval (dTime)
//                              --> Does NOT support COMOVING yet
//                RNG         : Random number generator
//                Efficiency  : Gas-to-star mass efficiency                                    (--> "SF_CREATE_STAR_MASS_EFF"      )
//                MinStarMass : Minimum star particle mass for the stochastical star formation (--> "SF_CREATE_STAR_MIN_STAR_MASS" )
//                TID         : OpenMP thread ID
```

src/StarFormation/SF_CreateStar_GetStarMass.cpp
```diff
@@ -0,0 +1,141 @@
...
+//-------------------------------------------------------------------------------------------------------
+real SF_CreateStar_GetStarMass_StochasticLoaclSchmidtLaw( const real GasDens, const real dv, const real dt, RandomNumber_t *RNG,
+                                                          const real Efficiency, const real MinStarMass, const int TID )
+{
+
+   real StarMass = -1.0;
+
+   const real   Coeff_FreeFall = SQRT( (32.0*NEWTON_G)/(3.0*M_PI) );
+   const real  _MinStarMass    = (real)1.0 / MinStarMass;
+   const real   Eff_times_dt   = Efficiency*dt;
```
hyschive:
```suggestion
   const real  Coeff_FreeFall = SQRT( (32.0*NEWTON_G)/(3.0*M_PI) );
   const real _MinStarMass    = (real)1.0 / MinStarMass;
   const real  Eff_times_dt   = Efficiency*dt;
```

src/TestProblem/Hydro/StarFormationThreshold/Init_TestProb_StarFormationThreshold.cpp
```diff
@@ -0,0 +1,438 @@
...
+      Aux_Error( ERROR_INFO, "OPT__FREEZE_FLUID must be enabled !!\n" );
+
+   for (int f=0; f<6; f++)
+   if ( OPT__BC_FLU[f] == BC_FLU_PERIODIC )
+      Aux_Error( ERROR_INFO, "do not use periodic BC (OPT__BC_FLU* = 1) for this test !!\n" );
+
+#  ifdef GRAVITY
+   if ( OPT__BC_POT == BC_POT_PERIODIC )
+      Aux_Error( ERROR_INFO, "do not use periodic BC (OPT__BC_POT = 1) for this test !!\n" );
+#  endif
```
hyschive: Why?

src/TestProblem/Hydro/StarFormationThreshold/Init_TestProb_StarFormationThreshold.cpp
```diff
@@ -0,0 +1,438 @@
...
+   SetParameter();
+
+
+// set the function pointers of various problem-specific routines
+   Init_Function_User_Ptr            = SetGridIC;
+   Par_Init_ByFunction_Ptr           = Par_Init_ByFunction_StarFormationThreshold;
+   Flu_ResetByUser_Func_Ptr          = Flu_ResetByUser_StarFormationThreshold;
+   Mis_GetTimeStep_User_Ptr          = Mis_GetTimeStep_StarFormationThreshold;
+#  ifdef SUPPORT_HDF5
+   Output_HDF5_InputTest_Ptr         = LoadInputTestProb;
```
hyschive:
```suggestion
   Init_Function_User_Ptr    = SetGridIC;
   Par_Init_ByFunction_Ptr   = Par_Init_ByFunction_StarFormationThreshold;
   Flu_ResetByUser_Func_Ptr  = Flu_ResetByUser_StarFormationThreshold;
   Mis_GetTimeStep_User_Ptr  = Mis_GetTimeStep_StarFormationThreshold;
#  ifdef SUPPORT_HDF5
   Output_HDF5_InputTest_Ptr = LoadInputTestProb;
```

example/test_problem/Hydro/StarFormationThreshold/Input__Parameter
```diff
@@ -0,0 +1,261 @@
...
+# 5. All dimensional variables should be set consistently with the code units (set by UNIT_L/M/T/V/D)
+# 6. For boolean options: 0/1 -> off/on
+# =================================================================================================================
+
+
+# simulation scale
+BOX_SIZE                      8.0         # box size along the longest side (in Mpc/h if COMOVING is adopted)
+NX0_TOT_X                     128         # number of base-level cells along x
+NX0_TOT_Y                     128         # number of base-level cells along y
+NX0_TOT_Z                     32          # number of base-level cells along z
```
hyschive, line 19:
```suggestion
NX0_TOT_Z                     16          # number of base-level cells along z
```

src/TestProblem/Hydro/StarFormationThreshold/Par_Init_ByFunction_StarFormationThreshold.cpp
```diff
@@ -0,0 +1,60 @@
...
+                                                 real_par *ParMass, real_par *ParPosX, real_par *ParPosY, real_par *ParPosZ,
+                                                 real_par *ParVelX, real_par *ParVelY, real_par *ParVelZ, real_par *ParTime,
+                                                 long_par *ParType, real_par *AllAttributeFlt[PAR_NATT_FLT_TOTAL],
+                                                 long_par *AllAttributeInt[PAR_NATT_INT_TOTAL] )
+{
+
+   if ( MPI_Rank == 0 )    Aux_Message( stdout, "%s ...\n", __FUNCTION__ );
+
+   if ( NPar_ThisRank != 0  ||  NPar_AllRank != 0 )
+      Aux_Error( ERROR_INFO, "total number of particles [%ld] != 0 !!\n", NPar_AllRank );
```
hyschive:
```suggestion


   if ( NPar_ThisRank != 0  ||  NPar_AllRank != 0 )
      Aux_Error( ERROR_INFO, "number of particles [this rank: %ld, total %ld] != 0 !!\n", NPar_ThisRank, NPar_AllRank );
```

example/test_problem/Hydro/StarFormationThreshold/README
```diff
@@ -0,0 +1,46 @@
...
+       [L] = kpc
+       [M] = 1.0e9 Msun
+       [T] = Myr
+    -->[V] = [L]/[T] ~ 9.8e2 km/s
+    -->[D] = [M]/[D]^3 ~ 6.8e-23 g/cm^3
+
+
+Note:
+========================================
+1. The initial conditions is a distribution of varying mass density and temperature
```
hyschive:
```suggestion
1. The initial conditions consist of a distribution of varying gas mass density and temperature
```

example/test_problem/Hydro/StarFormationThreshold/README
```diff
@@ -0,0 +1,46 @@
...
+   - Linearly varying log( mass density ) in the x direction
+   - Linearly varying log( temperature ) in the y direction
+   - No variation in the z direction
+
+2. Both the fluid and the particles are freezed and
```
hyschive:
```suggestion
2. Both the fluid and particles are frozen, and
```

example/test_problem/Hydro/StarFormationThreshold/README
```diff
@@ -0,0 +1,46 @@
...
+   the fluid is reset to the initial condition every step by default to keep a constant environment
```
hyschive:
```suggestion
   the fluid is reset to the initial conditions at every step by default to keep a constant environment
```

example/test_problem/Hydro/StarFormationThreshold/README
```diff
@@ -0,0 +1,46 @@
...
+   - One can also set OPT__RESET_FLUID = 0 to see how the fluid decreases and is converted into stars
```
hyschive:
```suggestion
   - One can also set OPT__RESET_FLUID = 0 to observe how the gas is depleted and converted into stars
```

example/test_problem/Hydro/StarFormationThreshold/README
```diff
@@ -0,0 +1,46 @@
...
+
+3. This is to test the star formation conditions that star should only form at the allowed region
```
hyschive:
```suggestion
3. This test is designed to verify that star formation occurs only in the allowed regions:
```

hyschive (commented):
@hsinhaoHHuang This PR looks good! I have only some minor comments. Please also check the following general suggestions:
- [ ] Update wiki
  - [ ] Update the accepted upper bound of `SF_CREATE_STAR_MASS_EFF`
  - [ ] Add the new parameter `SF_CREATE_STAR_MAX_GAS_JEANSL`
  - [ ] Mention the incompatibility between `SF_CREATE_STAR_SCHEME_DWARFGALAXY` and `JEANS_MIN_PRES`
- [ ] I suggest merging #349 and #351 first
  - [ ] Update this PR to support comoving coordinates after #349 is merged (at least for the star formation routines to be compatible with the changes introduced in #349; the `StarFormationThreshold` test can be updated in a separate PR if preferred)
  - [ ] Update this PR to support PUID after #351 is merged

example/test_problem/Hydro/StarFormationThreshold/Input__Parameter
```diff
@@ -0,0 +1,261 @@
...
+OPT__CORR_AFTER_ALL_SYNC     -1           # apply various corrections after all levels are synchronized (see "Flu_CorrAfterAllSync"):
+                                          # (-1=auto, 0=off, 1=every step, 2=before dump) [-1]
+OPT__NORMALIZE_PASSIVE        0           # ensure "sum(passive_scalar_density) == gas_density" [1]
+OPT__RESET_FLUID              1           # reset fluid variables after each update -> edit "Flu_ResetByUser.cpp" [0]
+OPT__FREEZE_FLUID             1           # do not evolve fluid at all [0]
+MIN_DENS                      0.0         # minimum mass density    (must >= 0.0) [0.0] ##HYDRO, MHD, and ELBDM ONLY##
+MIN_PRES                      0.0         # minimum pressure        (must >= 0.0) [0.0] ##HYDRO and MHD ONLY##
+MIN_EINT                      0.0         # minimum internal energy (must >= 0.0) [0.0] ##HYDRO and MHD ONLY##
+MIN_ENTR                      0.0         # minimum entropy         (must >= 0.0) [0.0] ##HYDRO and MHD ONLY##
+JEANS_MIN_PRES                1           # minimum pressure estimated from the Jeans length [0] ##HYDRO/MHD and GRAVITY ONLY##
```
hyschive: Should it be disabled?

src/Auxiliary/Aux_Check_Parameter.cpp
```diff
@@ -1853,9 +1853,18 @@ void Aux_Check_Parameter()
       Aux_Message( stderr, "WARNING : SF_CREATE_STAR_MIN_LEVEL (%d) > MAX_LEVEL (%d) --> no star particles will form !!\n",
                    SF_CREATE_STAR_MIN_LEVEL, MAX_LEVEL );
 
-   if ( SF_CREATE_STAR_SCHEME == SF_CREATE_STAR_SCHEME_AGORA  &&  !SF_CREATE_STAR_DET_RANDOM )
+   if ( SF_CREATE_STAR_SCHEME == SF_CREATE_STAR_SCHEME_DWARFGALAXY  &&  JEANS_MIN_PRES )
    {
-      Aux_Message( stderr, "WARNING : SF_CREATE_STAR_SCHEME == 1 will break bitwise reproducibility due to the \n" );
+      Aux_Message( stderr, "WARNING : SF_CREATE_STAR_SCHEME == %d is incompatible with JEANS_MIN_PRES !!\n",
```
hyschive, line 1924: Should we raise an error or warning?

include/Typedef.h
hsinhaoHHuang: Updated.

src/StarFormation/SF_CreateStar_Check.cpp
hsinhaoHHuang: Done.

src/StarFormation/SF_CreateStar_GeneralGalaxy.cpp
hsinhaoHHuang: I removed the whole section to simplify it.

src/StarFormation/SF_CreateStar_GeneralGalaxy.cpp
hsinhaoHHuang: I changed it to use `MHD_GetCellCenteredBEnergyInPatch()` to set Emag directly. It seems there is no need to allocate `Emag` as an array.

src/StarFormation/SF_CreateStar_Check.cpp
hsinhaoHHuang: In my understanding, this is the description for the AGORA scheme (`SF_CREATE_STAR_SCHEME_AGORA`), not for the `SF_CreateStar_Check_GasDensity()` itself. The references and the statement about the Jeans mass may be irrelevant to other schemes also using a density threshold.
For example, there may be another star formation scheme implemented in the future requiring that (1) density >= a threshold, (2) temperature <= a threshold, and (3) mass >= Jeans mass.
That is why I put it here.

src/StarFormation/SF_CreateStar_Check.cpp
hsinhaoHHuang: I intend to let each star formation scheme choose its own combination of multiple star formation criteria, which is common in the literature. The `break` is intended to separate different schemes. And, yes, usually they need to satisfy all the listed conditions in a scheme to form a star.
Meanwhile, the criteria-checking routine (e.g., density > a threshold) is general and can be shared by different schemes.

src/StarFormation/SF_CreateStar_GetStarMass.cpp
hsinhaoHHuang: Updated.

src/TestProblem/Hydro/StarFormationThreshold/Init_TestProb_StarFormationThreshold.cpp
hsinhaoHHuang: Removed. And I changed the default BC to periodic.

src/Auxiliary/Aux_Check_Parameter.cpp
hsinhaoHHuang: I am considering it.

Conceptually, the two are incompatible: if a pressure floor is added, the gas is stabilized against collapse and will not form stars. Numerically, however, they can coexist in the same run without causing a problem. It seems `JEANS_MIN_PRES` is not added inside the star formation check routine.

Is it possible that a user would want both treatments applied to the same Jeans-unresolved cell at MAX_LEVEL: adding `JEANS_MIN_PRES` to stabilize the fluid in the hydro solver, while simultaneously letting the star formation solver convert that gas into stars? (Both are sub-grid treatments.) This seems similar to the situation with density-based star formation models. If this is the case, perhaps a warning is enough.

hsinhaoHHuang (commented):
I have updated the star formation to support comoving coordinates. I added a new test in `StarFormationThreshold` test problem (see its `README`): Stars form in an environment of constant physical density and temperature in comoving evolution. The criteria check of physical density and Jeans length works. The resulting star formation rates and accumulated star masses can match the expectation.

hsinhaoHHuang:
To avoid future confusion, I am updating the description of `SF_CREATE_STAR_MIN_GAS_DENS` from `HI count/cm^3` to `m_H/cm^3`.

I think by expressing it as a mass density unit, this can be easily understood as how it will be converted, without assuming the gas composition (the term "HI count" may not be strictly correct). And it is easier to correspond to how the community defines the density threshold (some use hydrogen number density while others use total number density).

- In the GADGET-4 [document](https://wwwmpa.mpa-garching.mpg.de/gadget4/gadget4_manual.pdf), the parameter `CritPhysDensity` means "The critical physical hydrogen number density in cm^(-3) above which star formation is allowed." In their code, it is converted internally as `All.PhysDensThresh = All.CritPhysDensity * PROTONMASS / HYDROGEN_MASSFRAC / All.UnitDensity_in_cgs;` They are using the number density of hydrogen atoms.

- In the ENZO [code](https://github.com/enzo-project/enzo-dev/blob/main/src/enzo/Grid_StarParticleHandler.C#L1076-L1080), the parameter `StarMakerOverDensityThreshold` is in "particles/cc" and is converted internally as `float odthresh = StarMakerOverDensityThreshold * mh /  DensityUnits;` (I actually don't know why there is no mean molecular weight multiplied).

- In Goldbaum et al. 2015, $\rho_{\rm thresh} = \mu m_h n_{\rm thresh} = 50  {\rm cm}^{-3}$ with $\mu = 1.4$.

- In Kim et al. 2016, $n_{\rm H, thres} = 10 {\rm cm}^{-3} = \rho_{\rm gas, thres}/m_{\rm H}$.

hyschive:
@hsinhaoHHuang Could you resolve the conflicts? Thanks.

example/test_problem/Template/Input__Parameter
```diff
@@ -231,11 +231,12 @@ CHE_GPU_NPGROUP              -1           # number of patch groups sent into the
 
 
 # star formation (STAR_FORMATION only)
-SF_CREATE_STAR_SCHEME         0           # star formation schemes (0=off, 1=AGORA) [0]
+SF_CREATE_STAR_SCHEME         0           # star formation schemes (0=off, 1=AGORA, 2=DwarfGalaxy) [0]
```
hyschive, line 245: Just one quick comment. Should we keep the names `AGORA` and `DwarfGalaxy`, or switch to something more intuitive to indicate that `1` and `2` are based on density and Jeans length, respectively?

hsinhaoHHuang: I agree names indicating they are based on density and Jeans length would be much easier to understand.

Yet, I have trouble naming it.
One of my concerns is that a star formation 'scheme' contains more than the adopted threshold, and it may be difficult to express its features in its name
- I can imagine there are some star formation schemes with combined star formation criteria (e.g., high density + low temperature + converging velocity + gravitationally bound + ...).
- And there can be some different ways to decide the particle mass instead of the KS relation (e.g., the sink particle scheme in the other PR)
- There may be some different methods regarding the stochastic or deterministic spawning.

That is probably why I ended up following `AGORA` and using the project name as the scheme name.


I will look for some inspiration.

hyschive: @hsinhaoHHuang I see. It just that `AGORA` and `DwarfGalaxy` are not very intuitive to me. Let me know if you have any suggestions.

src/StarFormation/SF_CreateStar_GetStarMass.cpp
```diff
@@ -0,0 +1,151 @@
...
+//                dv               : Cell volume
+//                dt               : Time interval to advance solution
+//                RNG              : Random number generator
+//                Efficiency       : Gas-to-star mass efficiency                                    (--> "SF_CREATE_STAR_MASS_EFF"      )
+//                MinStarMass      : Minimum star particle mass for the stochastical star formation (--> "SF_CREATE_STAR_MIN_STAR_MASS" )
+//                TID              : OpenMP thread ID
+//
+// Return      :  StarMass
+//-------------------------------------------------------------------------------------------------------
+real SF_CreateStar_GetStarMass_StochasticLoaclSchmidtLaw( const real GasDens, const real CosmoScaleFactor, const real dv, const real dt, RandomNumber_t *RNG,
```
hyschive, line 77:
```suggestion
real SF_CreateStar_GetStarMass_StochasticLocalSchmidtLaw( const real GasDens, const real CosmoScaleFactor, const real dv, const real dt, RandomNumber_t *RNG,
```

example/test_problem/Template/Input__Parameter
hsinhaoHHuang: I would propose the following name changes
```
AGORA       -> ConstEffPerFreeFallTime_StochasticSpawning_HighGasDensity
DwarfGalaxy -> ConstEffPerFreeFallTime_StochasticSpawning_UnresolvedJeansLength
```
which follow (Star Formation Rate)_(Star Particle Spawning Approach)_(Star Formation Threshold).

Do you think they are too long?

hyschive: I think they are indeed a bit too long.

How about the following approach?
- Keep the _general_ option `SF_CREATE_STAR_SCHEME = SF_CREATE_STAR_SCHEME_AGORA / SF_CREATE_STAR_SCHEME_DWARFGALAXY`.
  - But perhaps shorten `DWARFGALAXY` as `DWARF` for conciseness.
- Add separate runtime parameters to control _individual_ star-formation options. For example:
  - The star formation criteria in `SF_CreateStar_Check()`.
  - Star mass in `SF_CreateStar_GetStarMass()`.
  - [Optional] Star-formation efficiency in `SF_CreateStar_GetStarMass_StochasticLoaclSchmidtLaw()`.
- Different `SF_CREATE_STAR_SCHEME` options adopt different _default_ settings for these individual options, which could be overwritten by explicitly specifying them in `Input__Parameter`.
  - Clarify the default settings on wiki.
- We could also add `SF_CREATE_STAR_SCHEME = SF_CREATE_STAR_SCHEME_CUSTOMIZED`, for which all individual options must be specified manually.

This is just my two cents. Any alternative suggestions are welcome!

PR #381, opened 2024-12-03
new branch

YiGuangLee:
．Updated to latest version of GAMER from branch main
．Add my work of stellar feedback
．Create single supernova feedback test problem.

PR #434, opened 2025-02-14
Support feedback from resolved Type II supernovae

hsinhaoHHuang:
## Goal 
- To support the Type II SNe feedback for isolated galaxy simulations.

## Changes
- Generalize the feedback to the leaf patches on levels < MAX_LEVEL
  - Change the option `FB_LEVEL` to `FB_MIN_LEVEL` to specify the minimum AMR level to apply feedback.
  - Add an array recording the non-leaf patches as an argument to the feedback routine. Non-local feedback should not be applied to the non-leaf patches in the feedback routine.
  - Particles in the non-leaf patches will not give feedback.
- Add feedback routine `FB_Resolved_SNeII()` and its runtime options `FB_RESOLVED_SNEII*`  in `Input__Parameter`
  - Suitable when there is sufficiently high star particle mass resolution to have at most one SN per particle (Maximum star particle mass*`FB_RESOLVED_SNEII_N_PER_MASS`<=1).
  - Suitable when there is sufficiently high grid resolution to resolve the Sedov phase of the blast wave caused by SN (so no kinetic/momentum feedback is needed).
  - Each newly formed star particle will be randomly sampled as an SN progenitor with a probability based on the star particle mass and the number of SN per stellar mass (`FB_RESOLVED_SNEII_N_PER_MASS`).
  - Each SN has the same fixed delay time (`FB_RESOLVED_SNEII_DELAY_TIME`). The explosion time is then decided during the star formation.
  - Each SN explosion deposits the thermal energy, mass, and metal (`FB_RESOLVED_SNEII_EJECT_ENGY`, `FB_RESOLVED_SNEII_EJECT_MASS`, `FB_RESOLVED_SNEII_EJECT_METAL`) into surrounding fluid cells where the particle is located.
  - If the mass of the cell where the particle is in is sufficient, then the feedback is applied to a single cell. Otherwise, the diameter of the feedback region will increase (1, 2,… FB_GHOST_SIZE+1) until the enclosed mass is higher than the given threshold `FB_RESOLVED_SNEII_MIN_M_GAS`. The feedback of mass and energy will be distributed uniformly in an inscribed sphere with the decided diameter.
  - Turn on `FB_RESOLVED_SNEII_RECORD` to record the information of the feedback event.
  - Must turn on `--feedback=true` and `--par_attribute_flt=1`(to have the `ParSNIITime` attribute) for `configure.py` during compilation.

- Add a new test problem, `Hydro/SNFeedbackBlastWave`
  - To simulate the effect when an SN explodes in a background ISM environment.
  - To see the blast wave caused by the SN explosion and compare it to `Hydro/BlastWave`.
  - To test the delay time of the SN explosion.
  - To test the amount of the deposited energy, mass, and metal by the SN.

- Add SNeII feedback in `Hydro/AGORA_IsolatedGalaxy`
  - To test the number of randomly sampled SNe and compare it to the number of stars.
  - To test the delay time of the SN by examining the recorded SNII explosion time.
  - This is for testing only and is not physically reasonable (the star particle mass is too large for this feedback scheme).

### Results of the SNeII feedback in `Hydro/SNFeedbackBlastWave`
  - Input
    - Background density = `1.0e-22 g/cm^3`
    - Background temperature = `10 K`
    - Grid resolution = `3.5 pc`
    - Star particle mass = `100 Msun` 
    - Star metal mass = `12.95 Msun`
    - Star particle velocity ~ `17 km/s` along the diagonal line of the box
    - SN delay time = `10 Myr`
    - SN deposited internal energy = `1.0e51 erg`
    - SN deposited mass = `15 Msun`
    - SN deposited metal = `3 Msun`
  - Slices of density
    - See the blast wave 
    Slice_z_density
  - Slices of temperature
    - See the total internal energy increases by `10^51 erg` at `t = 10 Myr`
    Slice_z_temperature
  - Slices of metal density
    - See the total metal mass increases by `3 Msun` at `t = 10 Myr` while the particle metal mass decreases by `3 Msun` and the particle total mass decreases by `15 Msun`.
    Slice_z_metal_density
  - Without the deposited mass and compared to the `Hydro/BlastWave` test problem (with the modified input files attached in `example/test_problem/Hydro/SNFeedbackBlastWave/ForBlastWave/`) --> very similar
    | SNFeedbackBlastWave | BlastWave           |
    |---------------------|---------------------|
    | image | image |
    | image | image |
  - Non-local feedback tests
    - background $\rho=10^{-24}\ \mathrm{g/cm^3}$, $dh=3.5\ \mathrm{pc}$, gas mass per cell = $0.63\ M_\odot$,target minimum gas mass = $10\ M_\odot$, actual enclosed mass = $12\ M_\odot$

       Slic_3
       image
    - background $\rho=10^{-24}\ \mathrm{g/cm^3}$, $dh=3.5\ \mathrm{pc}$, gas mass per cell = $0.63\ M_\odot$,target minimum gas mass = $1000\ M_\odot$, actual enclosed mass = $246\ M_\odot$ (limited by FB_GHOST_SIZE = 8)
       Slic_5
       image
    - background $\rho=10^{-24}\ \mathrm{g/cm^3}$, $dh=3.5\ \mathrm{pc}$, gas mass per cell = $0.63\ M_\odot$, target minimum gas mass = $10\ M_\odot$, actual enclosed mass = $5.1\ M_\odot$. When the particle is too close to the AMR coarse-fine boundary, the feedback region will be limited such that no feedback will be applied at the coarse level.
       Slic_7
       image
  - Non-MAX_LEVEL feedback
    - `FB_MIN_LEVEL = 3`, `MAX_LEVEL = 4`. The feedback can work as usual on the leaf-patches of Lv 3.
      Slic_1
      Data_000010_Slice_z_temperature_zoom-in_1
    - `FB_MIN_LEVEL = 3`, `MAX_LEVEL = 4`. The feedback will not cross the AMR coarse-fine boundary and is only applied to the leaf patches.
      Slic_2
      Data_000010_Slice_z_temperature_zoom-in_2

### Results of the SNeII feedback in `Hydro/AGORA_IsolatedGalaxy`
  - Input
    - `SF_CREATE_STAR_MIN_STAR_MASS` = `3.2e4 Msun`
    - `FB_RESOLVED_SNEII_N_PER_MASS` = `1.0e-5 per Msun`
    - `FB_RESOLVED_SNEII_DELAY_TIME` = `100.0 Myr`
    --> Note that the parameters for SNeII feedback are not physical and are for testing only.
  - Particle mass with exploded SN per year compared to the formed star mass per year, which are consistent. (Assumed each star particle has the same mass as the minimum star mass)
    image
  - The bitwise reproducibility with different MPI ranks and OpenMP threads are checked for non-local feedback.


hsinhaoHHuang:
I have updated the non-local feedback part, and this PR is ready for review again.

hsinhaoHHuang:
I have updated the feedback for leaf patches on non-MAX_LEVEL, and this PR is ready for review again.

barrytchiang (commented):
@hsinhaoHHuang I have finished a first full read of this PR at `6309e38a`, and thank you for pulling together this substantial amount of work! This round is based on source reading only; I will build it and run the test problems next and provide more detailed comments then.

## What I've checked that looks right

- The `NonLeaf[64]` index map in `FB_AdvanceDt()` step 5-4 agrees with `TABLE_GetSibPID_Delta()` and `FB_Aux_PatchLocalID()` in all 26 directions, faces, edges and corners alike. I also checked each delta list individually, the part I expected to be most error prone.
- I also agree with the structural choice of separating `Fluid_In` from `Fluid_Out`. The region growth decision reads the pre-feedback state, so it does not depend on the order in which particles are processed, while the deposits accumulate with `+=` so that several SNe or particles reaching the same cell add up correctly.
- The dual-energy variable is rebuilt per cell from the updated state through `Hydro_DensDual2Pres`, add `Eint`, `Hydro_DensPres2Dual`. That is the correct sequence, and more careful than what #381 did.
- The separate `FB_RNG` keeps the star formation draws unchanged whether feedback is on or off, and seeding it per patch from `LB_Idx` and the time is what lets the SN sampling be reproducible across different rank and thread counts.
- `FB_RNG` is allocated unconditionally in `FB_Init()`, so the reseed in the star formation routine is safe even with every feedback option disabled.
- The `SibPID_Delta` memory is now freed, and the patch lists are restricted to leaf patches on both the local and the sibling side.
- `SNFeedbackBlastWave` is self-contained: the yield table, the configure line, the comparison input files and the plot scripts are all included. README note 5 gives the exact parameter changes for each of the variants presented in the PR description, which made the review much easier.
- In the default setup the particle reaches the box center exactly when it explodes, which is a point shared by eight patch groups, so the deposit has to cross patch boundaries even in the default run rather than only in the variants of note 5.

## Main questions and suggestions

**1. `Aux_Check_Parameter()` checks the floating-point attribute count where it should check the integer one.** It requires `PAR_NATT_FLT_USER >= 1` when `FB_RESOLVED_SNEII` is on, but `ParSNIINxtE` is registered through `AddParticleAttributeInt()` in `Par_Init_Attribute()`, so what `FB_RESOLVED_SNEII` needs is one user-defined integer attribute, not a floating-point one, and the condition should be `PAR_NATT_INT_USER >= 1`. The current `SNFeedbackBlastWave` setup builds without error for a possibly wrong reason: `generate_make.sh` sets both `--par_attribute_flt=1` and `--par_attribute_int=1`, and the floating-point slot is there for the `ParMetalFrac` attribute that this test problem adds in its own initializer. Configure with `--par_attribute_int=1` alone and the build is correct but aborts here; configure with `--par_attribute_flt=1` alone and it passes here, then fails later in `Par_Init_Attribute()` with an attribute-count message that never mentions `FB_RESOLVED_SNEII`. This echoes Copilot's comment 3. The same floating-point wording appears in the wiki Restriction, the `Template` and AGORA preset comments, `plot_SN_feedback_rate.py` and the PR description, all of them left over from the earlier `ParSNIITime` design, so one sweep for that name would settle Copilot 3, 6, 12 and 14 together.
    
**2. The ceiling on the number of SNe per particle is derived from the minimum star particle mass.** `FB_Init_Resolved_SNeII()` sets `maxNumSNeIIPerSPar = MIN( ceil(2*SF_CREATE_STAR_MIN_STAR_MASS*FB_RESOLVED_SNEII_N_PER_MASS), FB_SNII_NXTE_SEPDIGIT )` and loads one yield table per count up to that value. The sampling in `SF_CreateStar_AGORA()`, however, caps at `ceil(2*mean)` computed from each particle's actual `StarMass`, and `StarMass` is never below `SF_CREATE_STAR_MIN_STAR_MASS`. It equals the minimum in one case only: when `GasMass*StarMFrac` comes out below the minimum, the routine forms a particle of exactly `SF_CREATE_STAR_MIN_STAR_MASS` with probability `StarMass/MinStarMass`. In every other case the particle takes the full `GasMass*StarMFrac`, which is larger. So any particle formed above the minimum mass can be assigned more SNe than there are loaded tables, and the range check in the feedback routine then calls `Aux_Error`.

I do not think the ceiling can be derived from the largest particle mass that will actually occur, since that depends on the densest cell the run ever reaches and is not known at initialization. Two alternatives are worth considering here are: (1) add a runtime parameter for the maximum number of SNe per particle, load that many tables, and check the sampled count against it at creation time rather than at explosion time; (2) limit the sampled count at creation to the number of tables loaded, in which case it is worth stating what happens to the SNe removed by that limit, since they leave the SN budget and no record entry notes the reduction.

Two related minor points. `SF_CREATE_STAR_MIN_STAR_MASS` has a default of 0, which makes the ceiling zero, so no table is loaded at all. And `ParSNIINxtE` stores the total count and the next event index in a single integer as `N*FB_SNII_NXTE_SEPDIGIT + idx`, which fails at the largest count that integer can hold: with `N = 10000`, the value once the last SN has exploded is `10000*10000 + 10000`, and reading it back gives `N = 10001, idx = 0`. The cap should be `FB_SNII_NXTE_SEPDIGIT - 1`. This is Copilot 9, with the last point added.

**3. The capped Poisson draw biases the SN budget low.** `SampledNum = MIN( Poisson(mean), ceil(2*mean) )` puts all the probability above the cap onto the cap itself, so the expected count comes out below `mean`. For `mean <= 0.5` the cap is 1 and the expectation becomes `1 - exp(-mean)`, which is a 21% deficit as `mean` approaches 0.5 from below and a 14% deficit for the AGORA low-resolution preset as configured (`3.2e4 Msun * 1.0e-5 per Msun = 0.32`). At `mean = 1` the cap is 2 and the deficit is 10%. The cap itself is the cause, so the practical fix is to make it generous rather than `2*mean`, which would also fit with making the maximum a parameter as in item 2. One thing to avoid: re-drawing until the value falls within the cap does not help, because that gives the Poisson conditioned on `X <= cap`, whose mean is lower still (0.80 against 0.90 at `mean = 1`). If the truncation is intended, saying so in the comment would help, since the SN rate is the quantity the AGORA test is meant to verify.

**4. Ejecta momentum is deposited but the corresponding kinetic energy is not.** The deposit adds `SNII_DepositedMass * par_vel[]` to `MOMX/Y/Z` and only `SNII_DepositedEnergy` to `ENGY`, so the cell gains momentum and mass without the energy that carrying them requires. Writing `dEkin` for the resulting change in `|p|^2/(2*rho)`, the internal energy implied by the conserved variables changes by `dE - dEkin`, while `DUAL` is built in the same loop from `flu_Eint + dE`, so the two descriptions of the same cell end up apart by exactly `dEkin`. The size is small for realistic SN parameters and I do not think it affects the test problems here, but it grows with the particle velocity, and in a dual-energy run the discrepancy is between `DUAL` and `ENGY` rather than a pure conservation error. This is Copilot 5; the quantity is the change in the cell kinetic energy rather than the bulk kinetic energy of the ejecta alone.

**5. When a deposit region is cut short, the record file does not say why.** The growth loop stops early for two different reasons: the region reached the largest diameter `FB_GHOST_SIZE` allows, or it ran into a coarse-fine boundary or a non-leaf patch. Both leave `fbFluidMass` below `FB_RESOLVED_SNEII_MIN_M_GAS`, and both appear in your own results (246 Msun against a 1000 Msun target for the first, 5.1 Msun against 10 Msun for the second). The achieved diameter and the enclosed mass are recorded, so the shortfall itself is visible, but the cause is not, and the two mean quite different things to a user: the first says the parameters are asking for more gas than `FB_GHOST_SIZE` can reach, the second says the particle sat next to a refinement boundary. One more column, or a flag, would separate them.

**6. Is `FB_MIN_LEVEL = 0` the intended default?** The old `FB_LEVEL` default of -1 mapped to `MAX_LEVEL`, while the new parameter defaults to 0, so an existing `FB_USER` input file that does not set the new name will now attempt feedback on every leaf level rather than only the finest, and an old `FB_LEVEL` line is silently unrecognized. `Grackle_Comoving/Input__Parameter` still uses the old name (Copilot 10), and a sweep of the other example input files for stragglers would be worthwhile.

**7. One yield table file per SN count might be demanding on the user end.** A run whose ceiling is N requires the files `N000001` through `N0000NN`, each with the matching number of rows and all mutually consistent. For the one-SN test problem this is fine, but for anything IMF-sampled this might not be manageable. Do you plan to include the table generator script under `tool/table_maker/`, in the same way #420 handles its table? That would also be a natural place to document the columns, including that `progenitor_mass` is carried for provenance and is not read by the code.

**8. Quantitative benchmarks of the result.** The qualitative comparisons against `Hydro/BlastWave` so far look great! I wonder whethr you plan on also providing quantitative comparisons that would be stronger than the side-by-side images, and `plot_diag.pgt` already produces the diagonal density profile, so overlaying the two profiles on one figure, or quoting the shock position at a few times, would be cleaner. It would also help to see the deposited energy, mass and metal budget as numbers rather than only in the slice figures, since `plot_slice_gas.py` already computes them. For the AGORA figure, a direct check that the sampled counts follow the intended distribution would also be highly informative, as the existing comparison (if I understand correctly) assumes every star particle has the minimum mass and so does not really test the sampling.

## Minor points

**9.** Both AGORA presets set `FB_RESOLVED_SNEII 1`, and `FB_Init_Resolved_SNeII()` aborts at startup if `FeedbackYieldTable_Resolved_SNeII_N000001` is not in the working directory, which for AGORA it is not; the only copy is under `SNFeedbackBlastWave/`. Since the README note added in this PR says SN feedback is not officially supported for that test problem, defaulting the option to 0 and describing how to turn it on may be cleaner than adding a table. This is Copilot 1 and 2.

**10.** The record buffer holds `maxNumRecSNeII = 100` events per OpenMP thread, and `Record_FB_Resolved_SNeII()` writes it out and resets the counters after each `FB_AdvanceDt(lv)`, so the limit is 100 events per thread per level per step. The 101st raises `Aux_Error`, which ends the run because of a diagnostic output; either writing the buffer out as soon as it fills, or allocating it large enough for the number of particles at hand, or counting the events that did not fit and reporting that count, would all avoid that. In the same function, the recorded `Par_ID` is the patch-local `par_idx` rather than `PAR_PUID`, which makes an event hard to trace back to a particle across steps; the file is opened, appended and closed by every rank in turn with an `MPI_Barrier` between ranks after each `FB_AdvanceDt()` even when there are no events to write; and the `fopen` return value is not checked. This echoes Copilot 7 and 8.

**11.** `SNII_DepositedMass = MIN( par_mass, SNII_AccumulatedMass )` correctly prevents a negative particle mass, but the next line divides the remaining metal mass by the updated `PAR_MASS`, which can now be exactly zero, giving a NaN `ParMetalFrac`; and because of the `par_mass <= 0.0` condition earlier in the loop the particle is then skipped from then on, so any remaining SNe and their energy are dropped with no record entry. No configuration in this PR can reach it, since the ejecta mass is always well below the particle mass there, but `SF_CREATE_STAR_MIN_STAR_MASS` may be set below the ejecta mass of a single SN, and the failure is silent. This is Copilot 4, with the second consequence added.

**12.** `FB_RESOLVED_SNEII` requires `RNG_CPP11`, but the `Aux_Error` saying so sits inside the star formation loop, so a run built otherwise only stops once the first star forms. Checking it in `Aux_Check_Parameter()` would catch it at startup.

**13.** Two notes in the `FB_Resolved_SNeII()` header describe the earlier design and now disagree with the code in the same file: note 11 says there is at most one SN per particle, which the Poisson sampling replaced, and note 12 says the energy is injected only into the cell containing the particle, which the region growth contradicts. Note 12 also has "thre" for "there". The comment block on the star formation side documents the at-most-one probability as well (Copilot 11).

**14.** The wiki carries the same two descriptions: the `FB_RESOLVED_SNEII` entry describes the single-progenitor sampling, and the `FB_RESOLVED_SNEII_N_PER_MASS` restriction says the value times the maximum star particle mass cannot exceed one. The growth description mentions the coarse-fine boundary but not non-leaf patches, which is now the other half of the rule. `AGORA_IsolatedGalaxy/README` note 5 refers to `FB_RESOLVED_SNEII_DELAY_TIME`, which no longer exists now that the explosion times come from the table.

**15.** Runs using `PAR_INIT = 3` will now need a `ParSNIINxtE` column in the particle file, since `Par_Init_ByFile()` reads all integer attributes. This fails loudly through the size check, so it is only a documentation gap, but it is worth a line in the wiki.

**16.** Typos: "Vaslues" in `AGORA_IsolatedGalaxy/README`, "nits" in the wiki, "compilatin" and "Numer" in `SNFeedbackBlastWave/README`, "fracion" in the blast wave initializer, "matals" in the table validation message in `FB_Resolved_SNeII.cpp`, and the README reference to `plot_diag.gpt` where the file is `plot_diag.pgt` (Copilot 13).

PR #438, opened 2025-02-20
Add scale factor to Jeans coefficient

YuriOku:
I would like to suggest adding the scale factor to the Jeans coefficient when COMOVING is enabled.

When COMOVING is on, $P_c = P_p a^5$, $\rho_c = \rho_p a^3$, $\Delta x_c = \Delta x_p / a$, where the $a$, $P$, $\rho$, and $\Delta x$ are the scale factor, pressure, density, and cell size, respectively, and the subscripts $c$ and $p$ denotes comoving and physical variables.

The Jeans minimum pressure floor is  $P_p =\frac{1}{\gamma\pi} N^2 G \rho_p^2 \Delta x_p^2$ on the physical coordinates. Applying the coordinate conversion above, we obtain $P_c =a \frac{1}{\gamma\pi} N^2 G \rho_c^2 \Delta x_c^2$. So, I added `Time[lv]` to `JeansMinPres_Coeff`.

Similarly, the Jeans length is $\lambda_p = \sqrt{\frac{\pi \gamma}{G} \frac{P_p}{\rho_p^2}}$ on the physical coordinates, and this becomes $\lambda_c = \sqrt{a^{-1} \frac{\pi \gamma}{G} \frac{P_c}{\rho_c^2}}$ on the comoving coordinates. So, I added `Time[lv]` in the denominator of  `JeansCoeff_Factor`.

### DIFF
--- a/src/Main/InvokeSolver.cpp
+++ b/src/Main/InvokeSolver.cpp
@@ -547,7 +547,7 @@ void Solver( const Solver_t TSolver, const int lv, const double TimeNew, const d
 #  if ( MODEL == HYDRO  &&  defined GRAVITY )
 #  ifdef COMOVING
    const real JeansMinPres_Coeff = ( JEANS_MIN_PRES ) ?
-                                   Time[lv]*NEWTON_G*SQR(JEANS_MIN_PRES_NCELL*amr->dh[JEANS_MIN_PRES_LEVEL])/(GAMMA*M_PI) : NULL_REAL;
+                                   TimeOld*NEWTON_G*SQR(JEANS_MIN_PRES_NCELL*amr->dh[JEANS_MIN_PRES_LEVEL])/(GAMMA*M_PI) : NULL_REAL;
 #  else
    const real JeansMinPres_Coeff = ( JEANS_MIN_PRES ) ?
                                    NEWTON_G*SQR(JEANS_MIN_PRES_NCELL*amr->dh[JEANS_MIN_PRES_LEVEL])/(GAMMA*M_PI) : NULL_REAL;
--- a/src/Main/InvokeSolver.cpp
+++ b/src/Main/InvokeSolver.cpp
@@ -545,8 +545,13 @@ void Solver( const Solver_t TSolver, const int lv, const double TimeNew, const d
 #  endif
 
 #  if ( MODEL == HYDRO  &&  defined GRAVITY )
+#  ifdef COMOVING
+   const real JeansMinPres_Coeff = ( JEANS_MIN_PRES ) ?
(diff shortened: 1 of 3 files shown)


hyschive:
@YuriOku Thanks for the suggestions! I agree. To support comoving coordinates, we simply need to replace $G$ with $aG$.

@hsinhaoHHuang will help review this PR.

src/Main/InvokeSolver.cpp
```diff
@@ -545,8 +545,13 @@ void Solver( const Solver_t TSolver, const int lv, const double TimeNew, const d
 #  endif
 
 #  if ( MODEL == HYDRO  &&  defined GRAVITY )
+#  ifdef COMOVING
+   const real JeansMinPres_Coeff = ( JEANS_MIN_PRES ) ?
+                                   Time[lv]*NEWTON_G*SQR(JEANS_MIN_PRES_NCELL*amr->dh[JEANS_MIN_PRES_LEVEL])/(GAMMA*M_PI) : NULL_REAL;
```
hsinhaoHHuang, line 550: How about using `TimeOld` or `TimeNew` in this function?

src/Refine/Flag_Real.cpp
```diff
@@ -76,8 +76,12 @@ void Flag_Real( const int lv, const UseLBFunc_t UseLBFunc )
 #  endif // # if ( MODEL == ELBDM )
 
 #  if ( MODEL == HYDRO  &&  defined GRAVITY )
+#  ifdef COMOVING
+   const real JeansCoeff_Factor       = M_PI/( Time[lv]*SQR(FlagTable_Jeans[lv])*NEWTON_G ); // flag if dh^2 > JeansCoeff_Factor*Gamma*Pres/Dens^2
```
hsinhaoHHuang, line 80: Could you also add a comment in Line.43 in `src/Refine/Flag_Check.cpp` about this (G -> G*a when `COMOVING`) ?

hsinhaoHHuang (approved):
@YuriOku @hyschive
I think it makes sense conceptually that we need to multiply the NEWTON_G by the scale factor in comoving simulations, so I just approve.
I haven't run any tests. Do you have any thoughts on how to test it?

I also checked other places with `NEWTON_G` and probably only the external potential routines (e.g., `src/SelfGravity/CPU_Gravity/CPU_ExtAcc_PointMass.cpp`) do not take  `COMOVING` into account. (`Coeff_FreeFall` in `SF_CreateStar_AGORA.cpp` is used in the physical frame, so there is no problem.)

src/Main/InvokeSolver.cpp
YuriOku: I have replaced `Time[lv]` with `TimeOld`.

src/Refine/Flag_Real.cpp
YuriOku: added a comment about the scale factor in the description of `Flag_Check()` function.

hyschive (approved):

PR #439, opened 2025-02-20
Apply dual-energy fix during interpolate iteration

YuriOku:
I would like to propose applying the dual-energy fix before performing the general check in the `Interpolate_Iterate` function. This helps avoid generating artificial shock after refining cells with a high Mach number.

I have attached snapshots from the Zeldovich pancake test problems with three different setups as an example. The first is the reference setup with `MAX_LEVEL=0` at z=2.5.
static

Next, I made the following changes in the input parameters and rerun using the code before making the update in this PR.
- `MAX_LEVEL=1` in `Input__Parameter`
- `INT_OPP_SIGN_0TH_ORDER=0` in `Input__Parameter`
- set threshold density to 2.0 at level 0 in `Input__Flag_Rho`

Then, fluctuations appear in the density profile, and the pressure in the center increases compared to the reference result, which is likely caused by the failure of the interpolation due to the negative internal energy.
before

The last figure shows the result of the updated code using the same input parameters. The dual energy fix avoids negative internal energy and achieves smooth interpolation, and the profiles agree with the reference result.
after

### DIFF
--- a/src/Interpolation/Interpolate.cpp
+++ b/src/Interpolation/Interpolate.cpp
@@ -339,6 +339,19 @@ void Interpolate_Iterate( real CData[], const int CSize[3], const int CStart[3],
 #        endif
 
 
+//       use dual-energy fix before the general check
+#        ifdef DUAL_ENERGY
+         const bool CheckMinPres_No = false;
+         const real UseDual2FixEngy = HUGE_NUMBER;
+         char dummy;    // we do not record the dual-energy status here
+
+         if ( !FData_is_Prim )
+            Hydro_DualEnergyFix( Temp[DENS], Temp[MOMX], Temp[MOMY], Temp[MOMZ], Temp[ENGY], Temp[DUAL],
+                                 dummy, EoS_AuxArray_Flt[1], EoS_AuxArray_Flt[2],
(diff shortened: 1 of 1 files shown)


hyschive:
@YuriOku Thanks for your contribution. The revised results look great!

Note that we already apply the dual-energy fix [here](https://github.com/gamer-project/gamer/blob/main/src/LoadBalance/LB_Refine_AllocateNewPatch.cpp#L1100-L1112) after grid refinement. So I assume your proposed modification works better because it avoids reducing the monotonic coefficient [here](https://github.com/gamer-project/gamer/blob/main/src/Interpolation/Interpolate.cpp#L295-L307). Do I understand it correctly?

@ChunYen-Chen Could you help review and test this PR?

YuriOku:
@hyschive Yes, in the second case, the general check fails because the internal energy is negative, and the monotonic coefficient is reduced there as you mentioned. Applying the double energy fix ensures a positive internal energy and avoids the monotonic coefficient decrease.

ChunYen-Chen (approved):
@YuriOku Thanks for the contribution. The PR looks good to me.

hyschive (approved):

PR #457, opened 2025-03-31
Update the magnetic energy when enforcing consistency in the interface magnetic field

ChunYen-Chen:
Currently, if `MINMOD_MAX_ITER` is enabled to reduce the minmod coefficient in each patch, the output of the magnetic field may become inconsistent due to variations in the minmod coefficient across nearby patches.

To address this issue, enabling `OPT__SAME_INTERFACE_B` enforces consistency of the magnetic field at patch interfaces. However, in the current implementation, the magnetic energy at the interface is not updated.

This PR ensures that magnetic energy is updated when `OPT__SAME_INTERFACE_B` is enabled.

### DIFF
--- a/src/Init/Init_GAMER.cpp
+++ b/src/Init/Init_GAMER.cpp
@@ -246,8 +246,9 @@ void Init_GAMER( int *argc, char ***argv )
 
 // ensure B field consistency on the shared interfaces between sibling patches
 #  if ( MODEL == HYDRO  &&  defined MHD )
+   const bool update_energy_No = false;
    if ( OPT__SAME_INTERFACE_B )
-   for (int lv=0; lv<NLEVEL; lv++)  MHD_SameInterfaceB( lv );
+   for (int lv=0; lv<NLEVEL; lv++)  MHD_SameInterfaceB( lv, update_energy_No );
 #  endif
 
 
--- a/src/Main/EvolveLevel.cpp
+++ b/src/Main/EvolveLevel.cpp
@@ -303,6 +303,24 @@ void EvolveLevel( const int lv, const double dTime_FaLv )
 
       } // if ( OPT__OVERLAP_MPI ) ... else ...
 
+#     if ( MODEL == HYDRO  &&  defined MHD )
+      if ( OPT__SAME_INTERFACE_B )
+      {
+         const bool update_energy_Yes = true;
+
+         TIMING_FUNC(   Buf_GetBufferData( lv, SaveSg_Flu, SaveSg_Mag, NULL_INT, DATA_GENERAL,
+                                           _ENGY, _MAG, Flu_ParaBuf, USELB_YES  ),
+                        Timer_GetBuf[lv][0],   TIMER_ON   );
+
+         TIMING_FUNC(   MHD_SameInterfaceB( lv, update_energy_Yes ),
+                        Timer_Flu_Advance[lv],   TIMER_ON   );
+
+         TIMING_FUNC(   Buf_GetBufferData( lv, SaveSg_Flu, SaveSg_Mag, NULL_INT, DATA_GENERAL,
+                                           _ENGY, _MAG, Flu_ParaBuf, USELB_YES  ),
+                        Timer_GetBuf[lv][0],   TIMER_ON   );
+      } // if ( OPT__SAME_INTERFACE_B )
+#     endif // #if ( MODEL == HYDRO  &&  defined MHD )
+
       amr->FluSg    [lv]             = SaveSg_Flu;
       amr->FluSgTime[lv][SaveSg_Flu] = TimeNew;
 #     ifdef MHD
--- a/src/Main/Main.cpp
+++ b/src/Main/Main.cpp
@@ -703,8 +703,9 @@ int main( int argc, char *argv[] )
          if ( OPT__VERBOSE  &&  MPI_Rank == 0 )
             Aux_Message( stdout, "   MHD_SameInterfaceB                       ... " );
 
+         const bool update_energy_Yes = true;
          for (int lv=0; lv<NLEVEL; lv++)
-         TIMING_FUNC(   MHD_SameInterfaceB( lv ),     Timer_Main[6],   TIMER_ON   );
+         TIMING_FUNC(   MHD_SameInterfaceB( lv, update_energy_Yes ),     Timer_Main[6],   TIMER_ON   );
 
          if ( OPT__VERBOSE  &&  MPI_Rank == 0 )
             Aux_Message( stdout, "done\n" );
--- a/src/Model_Hydro/MHD_SameInterfaceB.cpp
+++ b/src/Model_Hydro/MHD_SameInterfaceB.cpp
@@ -18,11 +18,12 @@
 //                5. Mainly for debugging purposes since this consistency should already be guaranteed
 //                   even when disabling OPT__SAME_INTERFACE_B
 //
-// Parameter   :  lv : AMR level
+// Parameter   :  lv            : AMR level
+//                update_energy : Whether to update the magnetic energy
 //
 // Return      :  Longitudinal B field on the interfaces between sibling patches
 //-------------------------------------------------------------------------------------------------------
-void MHD_SameInterfaceB( const int lv )
+void MHD_SameInterfaceB( const int lv, const bool update_energy )
 {
 
 // check
@@ -32,24 +33,78 @@ void MHD_SameInterfaceB( const int lv )
 #  endif
 
 
+   const int FluSg = amr->FluSg[lv];
    const int MagSg = amr->MagSg[lv];
 
 // iterate over all real and buffer patches
 #  pragma omp parallel for schedule( runtime )
    for (int PID=0; PID<amr->num[lv]; PID++)
    {
+      real ***Emag_old = NULL;
+
+      if ( update_energy )   Aux_AllocateArray3D( Emag_old, 3, PS1, PS1 );
+
 //    use the B field on the +x/+y/+z sides to overwrite that on the -x/-y/-z sides
 //    --> skip s=1/3/5
       for (int s=0; s<6; s+=2)
       {
+         const int d      = s/2;
          const int SibPID = amr->patch[0][lv][PID]->sibling[s];
 
+         if ( SibPID < 0 )   continue;
+
 //       some buffer patches may have magnetic == NULL --> skip them
-         if ( SibPID >= 0  &&
-              amr->patch[MagSg][lv][   PID]->magnetic != NULL  &&
-              amr->patch[MagSg][lv][SibPID]->magnetic != NULL     )
-            MHD_CopyPatchInterfaceBField( lv, PID, s, MagSg );
-      }
+         if ( amr->patch[MagSg][lv][   PID]->magnetic == NULL  ||
+              amr->patch[MagSg][lv][SibPID]->magnetic == NULL     )   continue;
+
+         if ( update_energy )
+         {
+            int ii, jj, kk;
+            real u[NCOMP_TOTAL];
+
+            for (int j=0; j<PS1; j++)
+            for (int i=0; i<PS1; i++)
+            {
+               switch ( d )
+               {
+                  case 0: ii = 0;  jj = i;  kk = j;  break;
+                  case 1: ii = j;  jj = 0;  kk = i;  break;
+                  case 2: ii = i;  jj = j;  kk = 0;  break;
+               }
+
+               for (int v=0; v<NCOMP_TOTAL; v++)   u[v] = amr->patch[FluSg][lv][PID]->fluid[v][kk][jj][ii];
+
+               Emag_old[d][j][i] = MHD_GetCellCenteredBEnergyInPatch( lv, PID, ii, jj, kk, MagSg );
+            }
+         } // if ( update_energy )
+
+         MHD_CopyPatchInterfaceBField( lv, SibPID, s+1, MagSg );
+
+         if ( update_energy )
+         {
+            int ii, jj, kk;
+            real u[NCOMP_TOTAL];
+
+            for (int j=0; j<PS1; j++)
+            for (int i=0; i<PS1; i++)
+            {
+               switch ( d )
+               {
+                  case 0: ii = 0;  jj = i;  kk = j;  break;
+                  case 1: ii = j;  jj = 0;  kk = i;  break;
+                  case 2: ii = i;  jj = j;  kk = 0;  break;
+               }
+
+               for (int v=0; v<NCOMP_TOTAL; v++)   u[v] = amr->patch[FluSg][lv][PID]->fluid[v][kk][jj][ii];
+
+               const real Emag = MHD_GetCellCenteredBEnergyInPatch( lv, PID, ii, jj, kk, MagSg );
+
+               amr->patch[FluSg][lv][PID]->fluid[ENGY][kk][jj][ii] += (Emag - Emag_old[d][j][i]);
+            }
+         } // if ( update_energy )
+      } // for (int s=0; s<6; s+=2)
+
+      if ( update_energy )   Aux_DeallocateArray3D( Emag_old );
    } // for (int PID=0; PID<amr->num[lv]; PID++)
 
 } // FUNCTION : MHD_SameInterfaceB
--- a/include/Prototype.h
+++ b/include/Prototype.h
@@ -557,7 +557,7 @@ void MHD_AllocateElectricArray( const int lv );
 void MHD_Aux_Check_InterfaceB( const char *comment );
 void MHD_Aux_Check_DivergenceB( const bool Verbose, const char *comment );
 void MHD_FixUp_Electric( const int lv );
-void MHD_SameInterfaceB( const int lv );
+void MHD_SameInterfaceB( const int lv, const bool update_energy );
 void MHD_CopyPatchInterfaceBField( const int lv, const int PID, const int SibID, const int MagSg );
 void MHD_BoundaryCondition_Outflow( real **Array, const int BC_Face, const int NVar, const int GhostSize,
                                     const int ArraySizeX, const int ArraySizeY, const int ArraySizeZ,


hfhsieh (commented):
@ChunYen-Chen @hyschive 
Before looking into the details, I would like to ask something:
- Since it's only during initialization that we do not need to update the magnetic energy in `MHD_SameInterfaceB()`, I think we can always perform the update, as it should not impact performance.
- Should we update the wiki documentation to note this behavior under `MINMOD_MAX_ITER`, or consider automatically enabling `OPT__SAME_INTERFACE_B` when `MINMOD_MAX_ITER > 0` in MHD simulations?

ChunYen-Chen:
@hfhsieh 
> Since it's only during initialization that we do not need to update the magnetic energy in MHD_SameInterfaceB(), I think we can always perform the update, as it should not impact performance.

Agreed. I’ll let you know once I’ve updated the code.

> Should we update the wiki documentation to note this behavior under `MINMOD_MAX_ITER`, or consider automatically enabling `OPT__SAME_INTERFACE_B` when `MINMOD_MAX_ITER > 0` in MHD simulations?

I will update the wiki document.
@hyschive Would you like to automatically enabling `OPT__SAME_INTERFACE_B` when `MINMOD_MAX_ITER > 0` in MHD simulations?

src/Model_Hydro/MHD_SameInterfaceB.cpp
```diff
@@ -32,24 +32,69 @@ void MHD_SameInterfaceB( const int lv )
...
+         for (int i=0; i<PS1; i++)
+         {
+            switch ( d )
+            {
+               case 0: ii = 0;  jj = i;  kk = j;  break;
+               case 1: ii = j;  jj = 0;  kk = i;  break;
+               case 2: ii = i;  jj = j;  kk = 0;  break;
+            }
+
+            for (int v=0; v<NCOMP_TOTAL; v++)   u[v] = amr->patch[FluSg][lv][PID]->fluid[v][kk][jj][ii];
```
hfhsieh, line 89: `u[v]` is not used. Thus, we can remove lines 60, 72, and 89.

src/Main/EvolveLevel.cpp
```diff
@@ -303,6 +303,22 @@ void EvolveLevel( const int lv, const double dTime_FaLv )
...
+                                           _ENGY, _MAG, Flu_ParaBuf, USELB_YES  ),
+                        Timer_GetBuf[lv][0],   TIMER_ON   );
+
+         TIMING_FUNC(   MHD_SameInterfaceB( lv ),
+                        Timer_Flu_Advance[lv],   TIMER_ON   );
+
+         TIMING_FUNC(   Buf_GetBufferData( lv, SaveSg_Flu, SaveSg_Mag, NULL_INT, DATA_GENERAL,
+                                           _ENGY, _MAG, Flu_ParaBuf, USELB_YES  ),
+                        Timer_GetBuf[lv][0],   TIMER_ON   );
+      } // if ( OPT__SAME_INTERFACE_B )
```
hfhsieh, line 319: Update this to `if ( OPT__SAME_INTERFACE_B == SAME_INTERFACE_B_YES )`

hfhsieh (commented):
Looks good to me. I have added some comments.

ChunYen-Chen:
@hfhsieh Thanks for the review. I have addressed the comments.

hfhsieh (approved):

hyschive:
@ChunYen-Chen Could you resolve the conflicts caused by the recent wiki updates? Thanks.

include/Typedef.h
```diff
@@ -561,6 +561,13 @@ const LoadParaMode_t
    LOAD_HDF5_OUTPUT = 2;
 
 
+// whether to enforce consistency of magnetic field at the patch interface
+typedef int SameInterfaceB_t;
+const SameInterfaceB_t
+   SAME_INTERFACE_B_NO  = 0,
+   SAME_INTERFACE_B_YES = 1;
```
hyschive: Add the `_DEFAULT` option to follow the convention of other parameters, such as `RSOLVER_1ST_DEFAULT`. Update other places using `OPT__SAME_INTERFACE_B` in this PR accordingly.
```suggestion
   SAME_INTERFACE_B_DEFAULT = -1,
   SAME_INTERFACE_B_NO      = 0,
   SAME_INTERFACE_B_YES     = 1;
```

src/Main/Main.cpp
```diff
@@ -123,7 +123,7 @@ bool                 OPT__FIXUP_ELECTRIC, OPT__CK_INTERFACE_B, OPT__OUTPUT_CC_MA
 bool                 OPT__OUTPUT_DIVMAG;
 int                  OPT__CK_DIVERGENCE_B;
 double               UNIT_B;
-bool                 OPT__SAME_INTERFACE_B;
+int                  OPT__SAME_INTERFACE_B;
```
hyschive:
```suggestion
SameInterfaceB_t     OPT__SAME_INTERFACE_B;
```

include/Global.h
```diff
@@ -133,7 +133,7 @@ extern bool             OPT__FIXUP_ELECTRIC, OPT__CK_INTERFACE_B, OPT__OUTPUT_CC
 extern bool             OPT__OUTPUT_DIVMAG;
 extern int              OPT__CK_DIVERGENCE_B;
 extern double           UNIT_B;
-extern bool             OPT__SAME_INTERFACE_B;
+extern int              OPT__SAME_INTERFACE_B;
```
hyschive:
```suggestion
extern SameInterfaceB_t OPT__SAME_INTERFACE_B;
```

src/Init/Init_Load_Parameter.cpp
```diff
@@ -320,7 +320,7 @@ void Init_Load_Parameter()
    ReadPara->Add( "DUAL_ENERGY_SWITCH",         &DUAL_ENERGY_SWITCH,              2.0e-2,          0.0,           NoMax_double   );
 #  endif
 #  ifdef MHD
-   ReadPara->Add( "OPT__SAME_INTERFACE_B",      &OPT__SAME_INTERFACE_B,           false,           Useless_bool,  Useless_bool   );
+   ReadPara->Add( "OPT__SAME_INTERFACE_B",      &OPT__SAME_INTERFACE_B,          -1,               NoMin_int,     1              );
```
hyschive:
```suggestion
   ReadPara->Add( "OPT__SAME_INTERFACE_B",      &OPT__SAME_INTERFACE_B,          -1,                      -1,     1              );
```

doc/wiki/Runtime-Parameters-related/[Runtime-Parameters]-Hydro.md
```diff
@@ -196,7 +197,7 @@ Only applicable when enabling the compilation option
 [[--dual | Installation:-Option-List#--dual]].
 
 <a name="OPT__SAME_INTERFACE_B"></a>
-* #### `OPT__SAME_INTERFACE_B` &ensp; (0=off, 1=on) &ensp; [0]
+* #### `OPT__SAME_INTERFACE_B` &ensp; (-1=set to default, 0=off, 1=on) &ensp; [depend]
```
Comment (hyschive) [inline doc/wiki/Runtime-Parameters-related/[Runtime-Parameters]-Hydro.md:200]: Mention that this option is enabled by default when `MINMOD_MAX_ITER` is enabled.

example/test_problem/Template/Input__Parameter
```diff
@@ -257,7 +257,7 @@ OPT__1ST_FLUX_CORR           -1           # correct unphysical results (defined
                                           # (<0=auto, 0=off, 1=3D, 2=3D+1D) [-1] ##MHM/MHM_RP/CTU ONLY##
 OPT__1ST_FLUX_CORR_SCHEME    -1           # Riemann solver for OPT__1ST_FLUX_CORR (-1=auto, 0=none, 1=Roe, 2=HLLC, 3=HLLE, 4=HLLD) [-1]
 DUAL_ENERGY_SWITCH            2.0e-2      # apply dual-energy if E_int/E_kin < DUAL_ENERGY_SWITCH [2.0e-2] ##DUAL_ENERGY ONLY##
-OPT__SAME_INTERFACE_B         0           # ensure B field consistency on the shared interfaces between sibling patches (for debugging) [0] ##MHD ONLY##
+OPT__SAME_INTERFACE_B        -1           # ensure B field consistency on the shared interfaces between sibling patches (for debugging) (<0=auto, 0=off, 1=on) [-1] ##MHD ONLY##
```
hyschive:
```suggestion
OPT__SAME_INTERFACE_B        -1           # ensure B field consistency on the shared interfaces between sibling patches (mainly for debugging) (-1=auto, 0=off, 1=on) [-1] ##MHD ONLY##
```

src/Init/Init_ResetParameter.cpp
```diff
@@ -324,6 +324,15 @@ void Init_ResetParameter()
 #  endif
 
 
+#  ifdef MHD
+   if ( OPT__SAME_INTERFACE_B < 0 )
```
hyschive:
```suggestion
   if ( OPT__SAME_INTERFACE_B == SAME_INTERFACE_B_DEFAULT )
```

src/Main/Main.cpp
```diff
@@ -698,7 +698,7 @@ int main( int argc, char *argv[] )
       TIMING_FUNC(   Flu_CorrAfterAllSync(),          Timer_Main[6],   TIMER_ON   );
 
 #     if ( MODEL == HYDRO  &&  defined MHD )
-      if ( OPT__SAME_INTERFACE_B )
+      if ( OPT__SAME_INTERFACE_B == SAME_INTERFACE_B_YES )
```
hyschive, line 704: Add the following comment
```
\\ it may be unnecessary to call `MHD_SameInterfaceB()` here, since we already call it every sub-step
\\ in EvolveLevel() after the fluid solver (unless something other than the fluid solver can also introduce
\\ inconsistencies in the B field, such as fix-up operations?).
```

src/Model_Hydro/MHD_SameInterfaceB.cpp
```diff
@@ -32,24 +32,64 @@ void MHD_SameInterfaceB( const int lv )
...
+            {
+               case 0: ii = 0;  jj = i;  kk = j;  break;
+               case 1: ii = j;  jj = 0;  kk = i;  break;
+               case 2: ii = i;  jj = j;  kk = 0;  break;
+            }
+
+            Emag_old[d][j][i] = MHD_GetCellCenteredBEnergyInPatch( lv, PID, ii, jj, kk, MagSg );
+         }
+
+         MHD_CopyPatchInterfaceBField( lv, SibPID, s+1, MagSg );
```
hyschive, line 87: Could you clarify the reason for switching from `MHD_CopyPatchInterfaceBField( lv, PID, s, MagSg );` to `MHD_CopyPatchInterfaceBField( lv, SibPID, s+1, MagSg );`? Is it to avoid an OpenMP race condition, since this change allows us to update the energy of `PID` instead of `SibPID`?

If we adopt `MHD_CopyPatchInterfaceBField( lv, SibPID, s+1, MagSg );`, please update the function header accordingly to `Always use the B field on the -x/-y/-z sides to overwrite that on the +x/+y/+z sides`.

src/Main/EvolveLevel.cpp
```diff
@@ -303,6 +303,22 @@ void EvolveLevel( const int lv, const double dTime_FaLv )
...
+         TIMING_FUNC(   Buf_GetBufferData( lv, SaveSg_Flu, SaveSg_Mag, NULL_INT, DATA_GENERAL,
+                                           _ENGY, _MAG, Flu_ParaBuf, USELB_YES  ),
+                        Timer_GetBuf[lv][0],   TIMER_ON   );
+
+         TIMING_FUNC(   MHD_SameInterfaceB( lv ),
+                        Timer_Flu_Advance[lv],   TIMER_ON   );
+
+         TIMING_FUNC(   Buf_GetBufferData( lv, SaveSg_Flu, SaveSg_Mag, NULL_INT, DATA_GENERAL,
+                                           _ENGY, _MAG, Flu_ParaBuf, USELB_YES  ),
+                        Timer_GetBuf[lv][0],   TIMER_ON   );
```
hyschive: I don't think it's necessary to update buffer patches here, since the buffer-patch data haven't been updated after invoking the fluid solver anyway. In fact, we may not even need to update buffer patches when calling `MHD_SameInterfaceB()` in `EvolveLevel()`. Please refer to the comments I suggest adding to the function header of `MHD_SameInterfaceB()`.

src/Main/EvolveLevel.cpp
```diff
@@ -303,6 +303,22 @@ void EvolveLevel( const int lv, const double dTime_FaLv )
...
       } // if ( OPT__OVERLAP_MPI ) ... else ...
 
+#     if ( MODEL == HYDRO  &&  defined MHD )
+      if ( OPT__SAME_INTERFACE_B == SAME_INTERFACE_B_YES )
+      {
+         TIMING_FUNC(   Buf_GetBufferData( lv, SaveSg_Flu, SaveSg_Mag, NULL_INT, DATA_GENERAL,
+                                           _ENGY, _MAG, Flu_ParaBuf, USELB_YES  ),
+                        Timer_GetBuf[lv][0],   TIMER_ON   );
+
+         TIMING_FUNC(   MHD_SameInterfaceB( lv ),
```
hyschive: I think it's a bug since the original code updates the incorrect `Sg`. Do you agree?
```suggestion
         TIMING_FUNC(   MHD_SameInterfaceB( lv, SaveSg_Flu, SaveSg_Mag ),
```

src/Init/Init_GAMER.cpp
```diff
@@ -246,7 +246,7 @@ void Init_GAMER( int *argc, char ***argv )
 
 // ensure B field consistency on the shared interfaces between sibling patches
 #  if ( MODEL == HYDRO  &&  defined MHD )
-   if ( OPT__SAME_INTERFACE_B )
+   if ( OPT__SAME_INTERFACE_B == SAME_INTERFACE_B_YES )
    for (int lv=0; lv<NLEVEL; lv++)  MHD_SameInterfaceB( lv );
```
hyschive: See my comments in `EvolveLevel()` and `MHD_SameInterfaceB()`
```suggestion
   for (int lv=0; lv<NLEVEL; lv++)  MHD_SameInterfaceB( lv, amr->FluSg[lv], amr->MagSg[lv] );
```

src/Main/Main.cpp
hyschive: See my comments in `EvolveLevel()` and `MHD_SameInterfaceB()`
```suggestion
         TIMING_FUNC(   MHD_SameInterfaceB( lv, amr->FluSg[lv], amr->MagSg[lv] ),
                                                      Timer_Main[6],   TIMER_ON   );
```

src/Model_Hydro/MHD_SameInterfaceB.cpp
hyschive: See my comment in `EvolveLevel()` for why this change is necessary. If you agree, please document the new parameters in the function header and update `Prototype.h` as well. 
```suggestion
void MHD_SameInterfaceB( const int lv, , const int FluSg, const int MagSg )
```

src/Model_Hydro/MHD_SameInterfaceB.cpp
```diff
@@ -32,24 +32,64 @@ void MHD_SameInterfaceB( const int lv )
 #  endif
 
 
+   const int FluSg = amr->FluSg[lv];
    const int MagSg = amr->MagSg[lv];
```
hyschive:
```suggestion
```

src/Model_Hydro/MHD_SameInterfaceB.cpp
```diff
@@ -32,24 +32,64 @@ void MHD_SameInterfaceB( const int lv )
...
 
 // iterate over all real and buffer patches
 #  pragma omp parallel for schedule( runtime )
    for (int PID=0; PID<amr->num[lv]; PID++)
    {
+      real ***Emag_old = NULL;
+
+      Aux_AllocateArray3D( Emag_old, 3, PS1, PS1 );
+
```
hyschive: To improve performance, call `Aux_AllocateArray3D()` only once per thread.
```suggestion
// start of OpenMP parallel region
#  pragma omp parallel
   {  
      real ***Emag_old = NULL;

      Aux_AllocateArray3D( Emag_old, 3, PS1, PS1 );

// iterate over all real and buffer patches
#  pragma omp for schedule( runtime )
   for (int PID=0; PID<amr->num[lv]; PID++)
   {
```

src/Model_Hydro/MHD_SameInterfaceB.cpp
```diff
@@ -32,24 +32,64 @@ void MHD_SameInterfaceB( const int lv )
...
+            }
+
+            const real Emag = MHD_GetCellCenteredBEnergyInPatch( lv, PID, ii, jj, kk, MagSg );
+
+            amr->patch[FluSg][lv][PID]->fluid[ENGY][kk][jj][ii] += (Emag - Emag_old[d][j][i]);
+         }
+      } // for (int s=0; s<6; s+=2)
+
+      Aux_DeallocateArray3D( Emag_old );
    } // for (int PID=0; PID<amr->num[lv]; PID++)
```
hyschive:
```suggestion
   } // for (int PID=0; PID<amr->num[lv]; PID++)

   Aux_DeallocateArray3D( Emag_old );

   } // end of OpenMP parallel region
```

src/Model_Hydro/MHD_SameInterfaceB.cpp
hyschive, line 32:
```suggestion
//                6. The following two approaches are equivalent. We currently adopt approach (1).
//                   --> (1) Iterate over both real and buffer patches. In this case, there is no need to call
//                           Buf_GetBufferData() afterward to exchange buffer-patch data.
//                       (2) Iterate over real patches only. In this case, we must call Buf_GetBufferData()
//                           afterward to exchange Flu_ParaBuf buffer-patch data for both the magnetic field and energy
//                7. Possible optimizations:
//                   --> (1) When calling MHD_SameInterfaceB() in EvolveLevel(), iterate over real patches only,
//                           since Buf_GetBufferData() is always called in EvolveLevel() anyway.
//                       (2) Related to the above, when calling Buf_GetBufferData() before MHD_SameInterfaceB()
//                           in EvolveLevel(), we could exchange only 0 instead of Flu_ParaBuf data for the
//                           magnetic field and skip exchanging the energy, since buffer patches do not need to be
//                           updated inside MHD_SameInterfaceB() here.
//
```

hyschive (changes requested):
@ChunYen-Chen Thanks for the contribution! Please let me know if you agree with the comments I've added. Also, I've filed a potentially related issue (#513).

hyschive:
@ChunYen-Chen Is it ready for another review?

src/Model_Hydro/MHD_SameInterfaceB.cpp
ChunYen-Chen: > Is it to avoid an OpenMP race condition, since this change allows us to update the energy of `PID` instead of `SibPID`?

Yes, since the original code updates neighboring patch data, there is a chance that two threads update the same patch corner cell simultaneously.

I have updated the comment.

hyschive (approved):

PR #469, opened 2025-06-05
Fftw.cpp malloc size fix

xuanweishan:
## Summary:
This PR reduce the virtual memory requirement of gamer with FFTW3.
 
### Init/Init_FFTW.cpp
1. Add `fftw_mpi_local_size_3d_transposed()` to `ComputePaddedTotalSize()` and `ComputeTotalSize()` for mpi supported situation to enhance the estimation of memory requirement.
2. Original `MemUnit` estimation is not proper, it can't decrease the memory usage per rank with mpi.  So we replace the value `MemUnit` from `amr->NPatchComma[0][1]*PS1` to `amr->NPatchComma[0][1]/MPI_NRank`
3. Add variable check to the `MemUnit` related `malloc` and `realloc` to prevent error in after usage.

## Tests:
- [x] `SERIAL` support
- [x] `OPT__FFTW_STARTUP = 1`  support

### DIFF
--- a/src/Init/Init_FFTW.cpp
+++ b/src/Init/Init_FFTW.cpp
@@ -54,7 +54,7 @@ size_t ComputeTotalSize(int* size) {
 #  else
    mpi_index_int local_nz, local_z_start, local_ny_after_transpose, local_y_start_after_transpose;
 
-   return (size_t)fftw_mpi_local_size_3d_transposed(size[2], size[1], size[0]),
+   return (size_t)fftw_mpi_local_size_3d_transposed(size[2], size[1], size[0],
                                                     MPI_COMM_WORLD, &local_nz, &local_z_start, &local_ny_after_transpose,
                                                     &local_y_start_after_transpose);
 #  endif


src/Init/Init_FFTW.cpp
```diff
@@ -26,8 +26,19 @@ gramfe_fftw::complex_plan_1d FFTW_Plan_ExtPsi, FFTW_Plan_ExtPsi_Inv;   // ExtPsi
 //
 // Return      :  length of array that is large enough to store FFT input and output
 //-------------------------------------------------------------------------------------------------------
+//int ComputePaddedTotalSize(int* size) {
+//   return 2*(size[0]/2+1)*size[1]*size[2];
+//} // FUNCTION : ComputePaddedTotalSize
 size_t ComputePaddedTotalSize(int* size) {
+#  if SERIAL
```
koarakawaii: maybe `#  ifdef SERIAL`?

src/Init/Init_FFTW.cpp
```diff
@@ -26,8 +26,19 @@ gramfe_fftw::complex_plan_1d FFTW_Plan_ExtPsi, FFTW_Plan_ExtPsi_Inv;   // ExtPsi
 //
 // Return      :  length of array that is large enough to store FFT input and output
 //-------------------------------------------------------------------------------------------------------
+//int ComputePaddedTotalSize(int* size) {
+//   return 2*(size[0]/2+1)*size[1]*size[2];
+//} // FUNCTION : ComputePaddedTotalSize
```
koarakawaii: Maybe can remove the comment~

src/Init/Init_FFTW.cpp
```diff
@@ -40,8 +51,19 @@ size_t ComputePaddedTotalSize(int* size) {
 //
 // Return      :  length of array that is large enough to store FFT input and output
 //-------------------------------------------------------------------------------------------------------
+//int ComputeTotalSize(int* size) {
+//   return size[0]*size[1]*size[2];
+//} // FUNCTION : ComputeTotalSize
```
koarakawaii: Maybe can remove the comment~

src/Init/Init_FFTW.cpp
```diff
@@ -40,8 +51,19 @@ size_t ComputePaddedTotalSize(int* size) {
...
 size_t ComputeTotalSize(int* size) {
+#  if SERIAL
```
koarakawaii: Maybe `#  ifdef SERIAL`?

src/Init/Init_FFTW.cpp
```diff
@@ -40,8 +51,19 @@ size_t ComputePaddedTotalSize(int* size) {
...
    return (size_t)size[0]*size[1]*size[2];
+#  else
+   mpi_index_int local_nz, local_z_start, local_ny_after_transpose, local_y_start_after_transpose;
+
+   return (size_t)fftw_mpi_local_size_3d_transposed(size[2], size[1], 2*(size[0]/2+1),
```
koarakawaii: Maybe `size[0]` instead of `2*(size[0]/2+1)` here~

src/Init/Init_FFTW.cpp
```diff
@@ -40,8 +51,19 @@ size_t ComputePaddedTotalSize(int* size) {
 //
 // Return      :  length of array that is large enough to store FFT input and output
 //-------------------------------------------------------------------------------------------------------
+//int ComputeTotalSize(int* size) {
+//   return size[0]*size[1]*size[2];
+//} // FUNCTION : ComputeTotalSize
 size_t ComputeTotalSize(int* size) {
```
koarakawaii, line 51: Since when using `MPI` the returned value here is the size required by each rank instead of the total size, so maybe we can rename it to something like `ComputeRequiredSize` to avoid misunderstanding~?

src/Init/Init_FFTW.cpp
```diff
@@ -26,8 +26,19 @@ gramfe_fftw::complex_plan_1d FFTW_Plan_ExtPsi, FFTW_Plan_ExtPsi_Inv;   // ExtPsi
 //
 // Return      :  length of array that is large enough to store FFT input and output
 //-------------------------------------------------------------------------------------------------------
+//int ComputePaddedTotalSize(int* size) {
+//   return 2*(size[0]/2+1)*size[1]*size[2];
+//} // FUNCTION : ComputePaddedTotalSize
 size_t ComputePaddedTotalSize(int* size) {
```
koarakawaii, line 29: Since when using `MPI` the returned value here is the padded size required by each rank instead of the padded total size, so maybe we can rename it to something like `ComputePaddedRequiredSize` to avoid misunderstanding~?

src/Init/Init_FFTW.cpp
```diff
@@ -363,6 +385,15 @@ void Patch2Slab( real *VarS, real *SendBuf_Var, real *RecvBuf_Var, long *SendBuf
...
       List_NSend_SIdx[r] = 0;
+
+      if ( List_PID[r] == NULL )
+          Aux_Error( ERROR_INFO, "List_PID[%d] is NULL on Rank %d !!\n", r, MPI_Rank );
+      if ( List_k[r] == NULL )
+          Aux_Error( ERROR_INFO, "List_k[%d] is NULL on Rank %d !!\n", r, MPI_Rank );
+      if ( TempBuf_SIdx[r] == NULL )
+          Aux_Error( ERROR_INFO, "TempBuf_SIdx[%d] is NULL on Rank %d !!\n", r, MPI_Rank );
+      if ( TempBuf_Var[r] == NULL )
+          Aux_Error( ERROR_INFO, "TempBuf_Var[%d] is NULL on Rank %d !!\n", r, MPI_Rank );
```
koarakawaii, line 390: Maybe can apply the same check for `List_PID`, `List_k`, `TempBuf_SIdx` and `TempBuf_Var` for pointers returned by `realloc` below as well~?

koarakawaii (commented):
Hi, @xuanweishan: some minor comments are added. Please have a look :smiley_cat: ~

hyschive (approved):
@xuanweishan @koarakawaii Looks good. Thanks for the quick fix!

PR #471, opened 2025-06-16
Generalize `HDF5_Output_t` 

ChunYen-Chen:
This PR generalizes the `HDF5_Output_t` structure and simplifies the procedure of adding a new parameter in the HDF5 snapshot.

## New
1. `HDF5_Output_t` can store array data.

## Optimization
1. Replace `KeyInfo_t`, `Makefile_t`, `SymConst_t`, and `InputPara_t` to `HDF5_Output_t`
2. Simplify the procedure of adding a new parameter in the HDF5 snapshot.
3. Combine the restart from HDF5 code to `Output_DumpData_Total_HDF5.cpp`.

## Others
1. Force the `*label` (`FieldName*`) size to be `MAX_STRING` (512) in analysis tools.

## Test done
- [x] `GAMER_CompareData`, `GAMER_ExtractProfile`, and `GAMER_ExtractUniform`
- [x] Test problem: `Acoustic` -> Hydro
- [x] Test problem: `CDM_LSS` -> Hydro Comoving
- [x] Test problem: `CR_ShockTube` -> Passive scalar
- [x] Test problem: `MHD_LinearWave` -> MHD
- [x] Test problem: `Plummer` -> Particle
- [x] Test problem: `GaussianWavePacket` -> ELBDM
- [x] Test problem: `Solition` -> Gravity, Flag tables
- [x] Test problem: `LSS` -> ELBDM Comoving
- [x] Test problem: `LSS_Hybrid` -> ELBDM hybrid


include/HDF5_Typedef.h
```diff
@@ -1032,82 +163,588 @@ struct HDF5_Output_t
...
+
+      RS_Compare      [NField] = Compare;
+      RS_FatalNonexist[NField] = Fatal_Nonexist;
+      RS_FatalCompare [NField] = Fatal_Compr;
+      ArrSize         [NField] = (size_t)1;
+      ArrDim          [NField] = Dim;
+      ArrLen          [NField] = NULL;
+
+//    copy field value
+      Ptr[NField] = new T [ (int)ArrSize[NField] ];
```
HsunYeong, line 310:
```suggestion
      Ptr[NField] = new T [ ArrSize[NField] ];
```
Might be better to initialize `Ptr[NField]` as `new char` since it is deleted as `delete [] (char*)Ptr[i]`?

include/HDF5_Typedef.h
```diff
@@ -1032,82 +163,588 @@ struct HDF5_Output_t
...
+   void Copy2Dataset( char *dataset )
+   {
+
+      size_t offset = (size_t)0;
+
+      for (int i=0; i<NField; i++)
+      {
+         if ( Type[i] == TYPE_BOOL ) // bool is stored as int
+         {
+            int *temp = new int [ (int)ArrSize[i] ];
```
HsunYeong:
```suggestion
            int *temp = new int [ ArrSize[i] ];
```

include/HDF5_Typedef.h
```diff
@@ -1032,82 +163,588 @@ struct HDF5_Output_t
...
+
+            for (int j=0; j<(int)ArrSize[i]; j++)
```
HsunYeong:
```suggestion
            for (size_t j=0; j<ArrSize[i]; j++)
```

HsunYeong, line 438:
```suggestion
            int *temp = new int [ ArrSize[i] ];

            for (size_t j=0; j<ArrSize[i]; j++)
```
`(int)ArrSize[i]` might cause overflow.

include/HDF5_Typedef.h
```diff
@@ -1032,82 +163,588 @@ struct HDF5_Output_t
...
+      H5_Status       = H5Tclose( H5_TypeID_Load  );
+
+//    check size
+      if ( H5_FieldSize != ArrSize[FieldIdx] * TypeSize[FieldIdx]  &&  MPI_Rank == 0 )
+         Aux_Message( stdout, "\"%s\" comparison might be incomplete due to the size mismatch (RS: %zu Bytes != RT: %zu Bytes)!!\n",
+                      KeyName, H5_FieldSize, ArrSize[FieldIdx] * TypeSize[FieldIdx] );
+
+//    comparison
+      char KeyNameWithIdx[MAX_STRING];
+      for (int offset=0; offset<(int)ArrSize[FieldIdx]; offset++)
```
HsunYeong, line 574:
```suggestion
      for (size_t offset=0; offset<ArrSize[FieldIdx]; offset++)
```

include/HDF5_Typedef.h
```diff
@@ -1032,82 +163,588 @@ struct HDF5_Output_t
...
+            }
+         } // if ( offset*TypeSize[FieldIdx] >= MinSize )
+
+         sprintf( KeyNameWithIdx, "%s", KeyName );
+         int skip = (int)ArrSize[FieldIdx];
+         for (int d=0; d<ArrDim[FieldIdx]; d++)
+         {
+            skip /= ArrLen[FieldIdx][d];
+            int idx = (offset / skip) % ArrLen[FieldIdx][d];
+            snprintf( KeyNameWithIdx+strlen(KeyNameWithIdx), sizeof(KeyNameWithIdx)-strlen(KeyNameWithIdx), "[%d]", idx );
```
HsunYeong, line 601:
```suggestion
         size_t skip = ArrSize[FieldIdx];
         for (int d=0; d<ArrDim[FieldIdx]; d++)
         {
            skip /= ArrLen[FieldIdx][d];
            size_t idx = (offset / skip) % ArrLen[FieldIdx][d];
            snprintf( KeyNameWithIdx+strlen(KeyNameWithIdx), sizeof(KeyNameWithIdx)-strlen(KeyNameWithIdx), "[%zu]", idx );
```

src/Output/Output_DumpData_Total_HDF5.cpp
```diff
@@ -1653,2526 +1642,1554 @@ void Output_DumpData_Total_HDF5( const char *FileName )
...
+   HDF5_InputPara->Add( "Grackle_Verbose",        &GRACKLE_VERBOSE,         0, NULL, Compare_Yes, Fatal_No, Fatal_No );
+   HDF5_InputPara->Add( "Grackle_Cooling",        &GRACKLE_COOLING,         0, NULL, Compare_Yes, Fatal_No, Fatal_No );
+   HDF5_InputPara->Add( "Grackle_Primordial",     &GRACKLE_PRIMORDIAL,      0, NULL, Compare_Yes, Fatal_No, Fatal_No );
+   HDF5_InputPara->Add( "Grackle_Metal",          &GRACKLE_METAL,           0, NULL, Compare_Yes, Fatal_No, Fatal_No );
+   HDF5_InputPara->Add( "Grackle_UV",             &GRACKLE_UV,              0, NULL, Compare_Yes, Fatal_No, Fatal_No );
+   HDF5_InputPara->Add( "Grackle_CMB_Floor",      &GRACKLE_CMB_FLOOR,       0, NULL, Compare_Yes, Fatal_No, Fatal_No );
+   HDF5_InputPara->Add( "Grackle_PE_Heating",     &GRACKLE_PE_HEATING,      0, NULL, Compare_Yes, Fatal_No, Fatal_No );
+   HDF5_InputPara->Add( "Grackle_PE_HeatingRate", &GRACKLE_PE_HEATING_RATE, 0, NULL, Compare_Yes, Fatal_No, Fatal_No );
+   HDF5_InputPara->Add( "Grackle_CloudyTable",    &GRACKLE_CLOUDY_TABLE,    0, NULL, Compare_Yes, Fatal_No, Fatal_No );
```
HsunYeong, line 2822:
```suggestion
   HDF5_InputPara->Add( "Grackle_CloudyTable",     GRACKLE_CLOUDY_TABLE,    0, NULL, Compare_Yes, Fatal_No, Fatal_No );
```

src/Init/Init_ByRestart_HDF5.cpp
```diff
@@ -135,189 +93,117 @@ void Init_ByRestart_HDF5( const char *FileName )
...
-#     if ( ELBDM_SCHEME == ELBDM_HYBRID )
-      amr->use_wave_flag[lv] = KeyInfo.UseWaveScheme[lv];
-#     endif
+      GetFieldData( "ConRef",     H5_SetID_KeyInfo, H5_TypeID_KeyInfo, &ConRef[0],             NotAllocate );
+      ConRefInitialized = true;
    }
 
-#  ifdef PARTICLE
-   amr->Par->NPar_Active_AllRank = KeyInfo.Par_NPar;
-#  endif
```
HsunYeong, line 291:
```
#  ifdef PARTICLE
   GetFieldData( "Par_NPar", H5_SetID_KeyInfo, H5_TypeID_KeyInfo, &amr->Par->NPar_Active_AllRank, NotAllocate );
#  endif
```
It is required to reset `Par_NPar` since `PAR_NPAR` in `Input__Parameter` might not be consistent with `RESTART`.

src/Init/Init_ByRestart_HDF5.cpp
```diff
@@ -1536,1168 +1286,60 @@ void LoadOnePatch( const hid_t H5_FileID, const int lv, const int GID, const boo
...
-   InputPara_t RS;   // RS = ReStart
+// load
+   H5_TypeID_Field = H5Tget_member_type( H5_TypeID_Target, H5_FieldIdx );
+   H5_FieldSize    = H5Tget_size( H5_TypeID_Field );
 
-   if ( *EndT < 0.0 )
-   {
-      LoadField( "EndT",      EndT,    SID, TID, Fatal, NullPtr, -1, NonFatal );
+// allocate
+   if ( Allocate )   Ptr = new T [ (int)H5_FieldSize ];
```
HsunYeong, line 1329: Since `Ptr` is a local copy, the allocation will not be carried outside of this function unless we pass the pointer by reference. There doesn't seem to have any GetFieldData() call that use `Allocate = true`, do you think this is still necessary? Or maybe we can separate the allocation to a different function?

include/HDF5_Typedef.h
```diff
@@ -1032,82 +163,588 @@ struct HDF5_Output_t
...
+
+      for (int d=0; d<Dim; d++)
+      {
+         ArrLen [NField][d] = (hsize_t)Len[d];
+         ArrSize[NField]   *= (size_t )Len[d];
+      }
+
+//    copy field from the pointer
+      Ptr[NField] = new char [ (int)ArrSize[NField] * (int)TypeSize[NField] ];
+      memcpy( Ptr[NField], NewPtr, ArrSize[NField] * TypeSize[NField] );
```
HsunYeong, line 249:
```suggestion
      Ptr[NField] = new char [ ArrSize[NField] * TypeSize[NField] ];
      memcpy( Ptr[NField], NewPtr, ArrSize[NField] * TypeSize[NField] );
```

- I think c++ will convert the value in the bracket to `size_t` so no need to do the `int` cast here? From my testing on eureka, `size_t` is the same as `unsigned long`.
- When `Type[NField] == TYPE_BOOL`, `NewPtr` will have size `ArrSize[NField] * sizeof(bool)`, but `memcpy` will copy `ArrSize[NField] * sizeof(int)` bytes. Therefore I think it is better to initialize `Ptr[NField]` with `ArrSize[NField] * sizeof(bool)` and only copy the necessary bytes for `bool`'s case, what do you think?

include/HDF5_Typedef.h
```diff
@@ -1032,82 +163,588 @@ struct HDF5_Output_t
...
+      RS_Compare      [NField] = Compare;
+      RS_FatalNonexist[NField] = Fatal_Nonexist;
+      RS_FatalCompare [NField] = Fatal_Compr;
+      ArrSize         [NField] = (size_t)1;
+      ArrDim          [NField] = Dim;
+      ArrLen          [NField] = NULL;
+
+//    copy field value
+      Ptr[NField] = new T [ (int)ArrSize[NField] ];
+      memcpy( Ptr[NField], &KeyVal, ArrSize[NField] * TypeSize[NField] );
```
HsunYeong, line 311: For `bool` type, both `Ptr[NField]` and `KeyVal` have `sizeof(bool)`, but `sizeof(int)` bytes are copied here.

src/Output/Output_DumpData_Total_HDF5.cpp
```diff
@@ -1653,2526 +1642,1554 @@ void Output_DumpData_Total_HDF5( const char *FileName )
...
-   for (int v=0; v<PAR_NATT_FLT_TOTAL; v++)
+   const bool Compare_Flag_Rho     = ( OPT__FLAG_RHO          ) ? Compare_Yes : Compare_No;
+   const bool Compare_Flag_RhoGrad = ( OPT__FLAG_RHO_GRADIENT ) ? Compare_Yes : Compare_No;
+   const bool Compare_Flag_Lohner  = ( Opt__FlagLohner        ) ? Compare_Yes : Compare_No;
+   const bool Compare_Flag_Angular = ( OPT__FLAG_ANGULAR      ) ? Compare_Yes : Compare_No;
+   const bool Compare_Flag_Radial  = ( OPT__FLAG_RADIAL       ) ? Compare_Yes : Compare_No;
+   const bool Compare_Flag_User    = ( OPT__FLAG_USER         ) ? Compare_Yes : Compare_No;
+
+   int H5_Arr_User[1] = { OPT__FLAG_USER_NUM };
+   int H5_Arr_Lv  [1] = { (Load_RS) ? MAX_LEVEL : NLEVEL-1    };
```
HsunYeong, line 3080: This will cause warnings such as
```
"FlagTable_Rho" comparison might be incomplete due to the size mismatch
```
when `MAX_LEVEL` of RT < `NLEVEL-1` of RS during restart even if `Input__Flag_Rho` is the same. Do you think we should always compare `NLEVEL-1` data to suppress this warning?

src/Init/Init_ByRestart_HDF5.cpp
```diff
@@ -66,46 +63,8 @@ void Init_ByRestart_HDF5( const char *FileName )
 
    const bool    Fatal             = true;
    const bool NonFatal             = false;
```
HsunYeong, line 65: These two parameters don't seem to be used anymore.

include/HDF5_Typedef.h
```diff
@@ -1032,82 +163,588 @@ struct HDF5_Output_t
...
+                  Aux_Error( ERROR_INFO, "\"%30s\" : RESTART file (%s) != runtime (%s) !!\n",
+                             KeyNameWithIdx, (char*)FieldPtr+offset*MAX_STRING, (char*)Ptr[FieldIdx]+offset*MAX_STRING );
+                  return -2;
+               }
+               else
+               {
+                  Check_Pass = false;
+                  if ( MPI_Rank == 0 )
+                  Aux_Message( stderr, "WARNING : \"%30s\" : RESTART file (%s) != runtime (%s) !!\n",
+                               KeyNameWithIdx, (char*)FieldPtr+offset*MAX_STRING, (char*)Ptr[FieldIdx]+offset*MAX_STRING );
```
HsunYeong, line 647: Since older gamer use variable strings, if you restart from an older gamer snapshot, it will show
```
comparison might be incomplete due to the size mismatch (RS: 8 Bytes != RT: 512 Bytes)!!
```
first, then `RESTART file (%s)` here will print a random string. Affected parameters include `CodeVersion`, `GitBranch`, and `GitCommit`.

Maybe we can check the old snapshots in a separate block such as:
```
if (H5Tis_variable_str(H5_TypeID_Field))
{
   ...
}
H5Tclose( H5_TypeID_Field );
```
What do you think?

include/HDF5_Typedef.h
```diff
@@ -1032,82 +163,588 @@ struct HDF5_Output_t
...
+            case TYPE_LONG:   COMPARE_NON_STRING_TYPE( KeyNameWithIdx, FORMAT_LONG, Fatal_Compr,       *((long*  )FieldPtr+offset), *((long*  )Ptr[FieldIdx]+offset) ); break;
+            case TYPE_UINT:   COMPARE_NON_STRING_TYPE( KeyNameWithIdx, FORMAT_INT,  Fatal_Compr,       *((uint*  )FieldPtr+offset), *((uint*  )Ptr[FieldIdx]+offset) ); break;
+            case TYPE_ULONG:  COMPARE_NON_STRING_TYPE( KeyNameWithIdx, FORMAT_LONG, Fatal_Compr,       *((ulong* )FieldPtr+offset), *((ulong* )Ptr[FieldIdx]+offset) ); break;
+            case TYPE_BOOL:   COMPARE_NON_STRING_TYPE( KeyNameWithIdx, FORMAT_BOOL, Fatal_Compr, (bool)*((int*   )FieldPtr+offset), *((bool*  )Ptr[FieldIdx]+offset) ); break;
+            case TYPE_FLOAT:  COMPARE_NON_STRING_TYPE( KeyNameWithIdx, FORMAT_REAL, Fatal_Compr,       *((float* )FieldPtr+offset), *((float* )Ptr[FieldIdx]+offset) ); break;
+            case TYPE_DOUBLE: COMPARE_NON_STRING_TYPE( KeyNameWithIdx, FORMAT_REAL, Fatal_Compr,       *((double*)FieldPtr+offset), *((double*)Ptr[FieldIdx]+offset) ); break;
```
HsunYeong, line 633: If RS and RT have different data types, this will print the wrong values. For example, if RS has `FLOAT8=0` and RT has `FLOAT8=1`:
```
WARNING : "                        Float8" : RESTART file ( 0                   ) != runtime ( 1                   ) !!
"TinyNumber" comparison might be incomplete due to the size mismatch (RS: 4 Bytes != RT: 8 Bytes)!!
WARNING : "                    TinyNumber" : RESTART file ( 4.14452302922905e-317) != runtime ( 2.22507385850720e-308) !!
"HugeNumber" comparison might be incomplete due to the size mismatch (RS: 4 Bytes != RT: 8 Bytes)!!
WARNING : "                    HugeNumber" : RESTART file ( 1.05685337195934e-314) != runtime ( 1.79769313486232e+308) !!
"MaxError" comparison might be incomplete due to the size mismatch (RS: 4 Bytes != RT: 8 Bytes)!!
WARNING : "                      MaxError" : RESTART file ( 4.43665288467204e-315) != runtime ( 1.00000000000000e-14) !!
```
Maybe we can use something like:
```
case TYPE_FLOAT:
   if ( H5_FieldSize == sizeof(float) )
      COMPARE_NON_STRING_TYPE( KeyNameWithIdx, FORMAT_REAL, Fatal_Compr, *((float* )FieldPtr+offset), *((float* )Ptr[FieldIdx]+offset) );
   if ( H5_FieldSize == sizeof(double) )
      COMPARE_NON_STRING_TYPE( KeyNameWithIdx, FORMAT_REAL, Fatal_Compr, *((double*)FieldPtr+offset), *((float* )Ptr[FieldIdx]+offset) );
   break;
```
What do you think?

HsunYeong (commented):
@ChunYen-Chen Thank you for your fantastic PR! I've left some comments. Could you also update your gamer to the latest version since there are a few new parameters introduced.

src/Init/Init_ByRestart_HDF5.cpp
ChunYen-Chen: I modeled this after the `Aux_LoadTable()` function because `GetFieldData()` may need to load table-like data from an HDF5 file. So, I decided to keep this extensibility for now.

PR #481, opened 2025-09-16
Add alias expand to support alias command

### LINKED ISSUE #424: [Bug] Autocomplete alias command
Edge case: If user define `alias` or load some module in thier `.bashrc`, the sub-shell here will behave differently and might cause problem.
For example, define `alias python='python3'`.
Compare to `./configure.py` before, I'm not sure which one will be better.
Just put an edge case out.
I think that user will need to modify their python path at some place eventually. Similiar problem apply to all of the `generate_make.sh`.

_Originally posted by @technic960183 in https://github.com/gamer-project/gamer/pull/357#discussion_r1925377920_

xuanweishan:
# Bug fix #424
## Description
This pull request fix #424. Make user able to use autocomplete with python `alias` setting 

## Changes
1. Add a local variable `alias_expanded` in `/tool/config/config_autocomplete.sh` with definition:
    ```bash
    alias_expanded=$(alias -- "${COMP_WORDS[0]}" 2>/dev/null | sed -E "s/^alias ${COMP_WORDS[0]}='([^']+)'/\1/")
    ```
2. Add a condition for assigning variable  `configure_command`
    * If `alias_expanded` is not empty: `configure_command=${alias_expanded} ${COMP_WORDS[1]}`
    * If `alias_expanded` is empty:        `configure_command=${COMP_WORDS[0] ${COMP_WORDS[1]}}`

## Tests
- [x] `./configure [tab] [tab]`
- [x] `python configure [tab] [tab]` (w/ `alias` setting: `alias python="python3")
- [x] `python3 configure [tab] [tab]`

### DIFF
--- a/tool/config/config_autocomplete.sh
+++ b/tool/config/config_autocomplete.sh
@@ -25,7 +25,13 @@ __gamer_configure_autocomplete() {
     local configure_filename configure_command
     if [[ "${COMP_WORDS[0]}" == "python"* ]]; then
         configure_filename=${COMP_WORDS[1]}
-        configure_command="${COMP_WORDS[0]} ${COMP_WORDS[1]}"
+        local alias_expanded
+        alias_expanded=$(alias -- "${COMP_WORDS[0]}" 2>/dev/null | sed -E "s/^alias ${COMP_WORDS[0]}='([^']+)'/\1/")
+        if [[ -n "$alias_expanded" ]]; then
+            configure_command="${alias_expanded} ${configure_filename}"
+        else
+            configure_command="${COMP_WORDS[0]} ${COMP_WORDS[1]}"
+        fi
     else
         configure_filename=${COMP_WORDS[0]}
         configure_command="${COMP_WORDS[0]}"


tool/config/config_autocomplete.sh
```diff
@@ -25,7 +25,13 @@ __gamer_configure_autocomplete() {
     local configure_filename configure_command
     if [[ "${COMP_WORDS[0]}" == "python"* ]]; then
         configure_filename=${COMP_WORDS[1]}
-        configure_command="${COMP_WORDS[0]} ${COMP_WORDS[1]}"
+        local alias_expanded
+        alias_expanded=$(alias -- "${COMP_WORDS[0]}" 2>/dev/null | sed -E "s/^alias ${COMP_WORDS[0]}='([^']+)'/\1/")
```
technic960183: You may want to rename `alias_expanded` to `alias_def`, since the expansion catching will be move to the line with `eval` using `#alias ${COMP_WORDS[0]}=`.
```suggestion
        alias_expanded=$(alias -- "${COMP_WORDS[0]}" 2>/dev/null)
```

tool/config/config_autocomplete.sh
```diff
@@ -25,7 +25,13 @@ __gamer_configure_autocomplete() {
...
+        if [[ -n "$alias_expanded" ]]; then
+            configure_command="${alias_expanded} ${configure_filename}"
```
technic960183:
```suggestion
            eval "configure_command=( ${alias_expanded#alias ${COMP_WORDS[0]}=} ${configure_filename} )"
```

tool/config/config_autocomplete.sh
```diff
@@ -25,7 +25,13 @@ __gamer_configure_autocomplete() {
...
+        else
+            configure_command="${COMP_WORDS[0]} ${COMP_WORDS[1]}"
```
technic960183: Use an array to store `configure_command`.
```suggestion
            configure_command=("${COMP_WORDS[0]}" "${COMP_WORDS[1]}")
```

tool/config/config_autocomplete.sh
```diff
@@ -25,7 +25,13 @@ __gamer_configure_autocomplete() {
...
+        fi
     else
         configure_filename=${COMP_WORDS[0]}
         configure_command="${COMP_WORDS[0]}"
```
technic960183: Use an array to store `configure_command`.
```suggestion
        configure_command=("${COMP_WORDS[0]}")
```

xuanweishan:
@technic960183 
Thanks for the great suggestion. I've apply your suggestion in latest commit.

PR #488, opened 2025-11-09
Add GPU architecture BLACKWELL

koarakawaii:
* Some properties for `cudaDeviceProp` are deprecated in `CUDA 13.0`; I did not check all of them, but only modified those affecting the compilation. See https://docs.nvidia.com/cuda/cuda-runtime-api/structcudaDeviceProp.html#structcudaDeviceProp for details.
    * Can consider switching the [`cudaDeviceGetAttribute`](https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__DEVICE.html#group__CUDART__DEVICE_1gb22e8256592b836df9a9cc36c9db7151) API; but again, I did not check whether it covers all the properties we queried for `cudaDeviceProp`
* Support latest `cufftDx`
    * Modify macro `NONE` to `OPTION_NONE` in `GAMER` to be compatible with `cufftDx`'s `.hpp` file. 
* Add configure file `spock_intel_RTX5090.config` for RTX5090 on `spock26` and GB10 for `gx10`~, which uses private `NVIDIA HPC SDK` modules instead of public `CUDA` on the log-in node. The path should be directing to public `CUDA 13.0` when available, if we want to install it~.
    * ~`CUDA 12.1` also works for current driver version `580.95.05`, but I am not sure we want to fall back to `CUDA 12.1` if a more advanced version has been released.~
    * `cufftDx` on `gx10` is under private path.

### DIFF
--- a/configs/spark_gnu-arm_GB10.config
+++ b/configs/spark_gnu-arm_GB10.config
@@ -7,7 +7,7 @@ HDF5_PATH       /software/hdf5/1.10.6-gnu_13.3.0
 GRACKLE_PATH
 GSL_PATH        /software/gsl/2.6-gnu-13.3.0
 LIBYT_PATH
-CUFFTDX_PATH    /work1/koarakawaii/cuFFTDx/nvidia-mathdx-25.12.0-cuda13/nvidia/mathdx/25.12
+CUFFTDX_PATH    /software/cuFFTDx/25.12
 
 # compilers
 CXX     g++


xuanweishan:
It seems good to me.
I tried tests on both `eureka` and `spock`:
- [x] `eureka` 2080super(cuda12)
- [x] `spock` 3080ti(cuda12) 
- [x] `spock` 5090(cuda13)
- [x] `spock` 3080ti(cuda13)(With modify config file for 3080 ti)
All works properly.
I think this PR is ready to merge after resolved the conflicts.

configs/spark_gnu-arm_GB10.config
```diff
@@ -0,0 +1,49 @@
+# NTU-spock
+CUDA_PATH       /usr/local/cuda
+FFTW2_PATH
+FFTW3_PATH      /software/fftw/3.3.10-gnu-13.3.0-openmpi-4.1.5-ucx_mt
+MPI_PATH        /software/openmpi/4.1.5-ucx_mt-gnu-13.3.0
+HDF5_PATH       /software/hdf5/1.10.6-gnu_13.3.0
+GRACKLE_PATH
+GSL_PATH        /software/gsl/2.6-gnu-13.3.0
+LIBYT_PATH
+CUFFTDX_PATH    /work1/koarakawaii/cuFFTDx/nvidia-mathdx-25.12.0-cuda13/nvidia/mathdx/25.12
```
xuanweishan:
```suggestion
CUFFTDX_PATH    /software/cuFFTDx/25.12
```
I've put the mathdx headers to `/software/cuFFTDx`

koarakawaii:
@xuanweishan Thank for all the testing~ I have updated the `cufftDx` path to the public one.

configs/spark_gnu-arm_GB10.config
koarakawaii: Updated~

Issue #465, opened 2025-04-30
MPI Barrier Synchronization Mismatch in InvokeSolver()


hsinhaoHHuang:
### 🔎 What happened?

When
(1) the Makefile option `TIMING_SOLVER` is on (`--timing_solver=true` in `generate_make.sh`),
(2) the runtime option `OPT__TIMING_BARRIER` in `Input__Parameter` is on, and
(3) the number of patches in different ranks is not balanced,
the program may hang.

The problem might come from the loop in `InvokeSolver(), src/Main/InvokeSolver.cpp`
```c++
   for (Disp=NPG_Max; Disp<NTotal; Disp+=NPG_Max)
   {

      ArrayID      = 1 - ArrayID;
      NPG[ArrayID] = ( NPG_Max < NTotal-Disp ) ? NPG_Max : NTotal-Disp;


//-------------------------------------------------------------------------------------------------------------
      TIMING_SYNC(   Preparation_Step( TSolver, lv, TimeNew, TimeOld, NPG[ArrayID], PID0_List+Disp, ArrayID, GlobalTree ),
                     Timer_Pre[lv][TSolver]  );
//-------------------------------------------------------------------------------------------------------------


//-------------------------------------------------------------------------------------------------------------
#     ifdef GPU
      CUAPI_Synchronize();
#     endif
//-------------------------------------------------------------------------------------------------------------


//-------------------------------------------------------------------------------------------------------------
      TIMING_SYNC(   Solver( TSolver, lv, TimeNew, TimeOld, NPG[ArrayID], ArrayID, dt, Poi_Coeff ),
                     Timer_Sol[lv][TSolver]  );
//-------------------------------------------------------------------------------------------------------------


//-------------------------------------------------------------------------------------------------------------
      TIMING_SYNC(   Closing_Step( TSolver, lv, SaveSg_Flu, SaveSg_Mag, SaveSg_Pot,
                     NPG[1-ArrayID], PID0_List+Disp-NPG_Max, 1-ArrayID, dt ),
                     Timer_Clo[lv][TSolver]  );
//-------------------------------------------------------------------------------------------------------------

   } // for (int Disp=NPG_Max; Disp<NTotal; Disp+=NPG_Max)
```
, where `TIMING_SYNC` is defined in `include/Timer.h`:
```c++
// macro for timing solvers
#if ( defined TIMING_SOLVER  &&  defined TIMING )

#  ifdef GPU
[...]

PR #489, opened 2025-11-10
Temporarily fix for `MPI Barrier` mismatch in `InvokeSolver`

koarakawaii:
* This is only a temporary fix for issue #465 ; the PR is filed for recording purpose only
    * Need not only to match the `TIMING_SYNC` for `Solver` and `Closing_Step`, but also mimic the exactly same preprocessor (`ifdef GRAVITY ...`, etc) in function `Preparation_Step` for initiating `TIMING_SYNC`. This makes the code hard to maintain. So I think **we should not really merge this**.
* If you agree, I'll close this PR @hyschive .


hyschive:
Replaced by #519

PR #492, opened 2025-11-14
Bugfix for Test Problem `ELBDM_DiskHeating`

koarakawaii:
* add preprocessor `#ifdef STORE_PAR_ACC`
 * disallow configuration with `store_par_acc = False`


hyschive:
@koarakawaii Thanks!

PR #315, opened 2024-05-22
Improve Par_EquilibriumIC

hsinhaoHHuang:
## Goal
- To resolve #188 
- To resolve #201 
- To resolve #202 
- To resolve #295 

## Changes
1. Adjust the user interface
    1. Move the part with reading the parameters from the input files into `TestProblem/Hydro/ParticleEquilibriumIC/Init_TestProb_Hydro_ParEqmIC.cpp` like other test problems.
    2. Use `setXXX()` as the interface to set the parameters for the cloud constructors from test problems, and only the necessary parameters are needed to be set.
    3. Support adding analytical external potential (only the Plummer potential now).
    4. Output the relative error of the total enclosed mass by comparing it to the analytical solution instead.
    5. Output the radial distribution and energy distribution for debugging.
    6. Separate the number of bins of the input density profile, input external potential, inner radial profiles, and inner energy distribution.

2. Adjust the calculation
    1.  Unify the processes for density profiles from the table and from the analytical models.
    2. Construct the radial profiles more consistently, from the density profile to the enclosed mass profile, the potential profile , and finally the drho_dPsi profile.
    3. Use the central difference for the differentiation instead of linear regression.
    4. Add the external potential before the usage of potential to be more consistent.
    5. Include the low-binding-energy (high-velocity) distribution function.
    7. Use linear extrapolation for the density profile at r = 0.
    8. Integrate the enclosed mass profile with a higher-order shell mass for the density table.
    9. Use interpolation to sample E from the probability distribution.
    10. Use interpolation to find the drho/dPsi for a given Psi.

3. Adjust the format
    1. Rename some variables and functions.
    2. Remove some no longer used variables and functions.
    3. Add more description.
    4. Adjust the coding styles.

## Verification Tests
1. Compare the distribution to the old branch with the default double-clouds setup. ->The results can roughly match the previous results.
    1. Type = NFW 

    | radial | energy |
    | ---------| ---------- |
    | image | image |

    2. Type = Table

   | radial | energy |
   | ---------| ---------- |
   | image| image|

2. Compare the distribution of the Hernquist model to the analytical solutions. -> The results can match analytical solutions.
image

3. Check the stability of a single NFW profile with the default setup. -> The results are stable.

| New | Old |
| -------| ------ |
| Proj| Proj|
| Prof| Prof|
|fig_ErrorDensityProfile |fig_ErrorDensityProfile |
[...]

Issue #295, opened 2024-04-08
Improve Par_EquilibriumIC.cpp


hyschive:
- Make it more user-friendly
- #202 
- #201 
- Update all test problems using this tool

jzuhone:
@hyschive i am not sure if `ClusterMerger` is in mind under the “update all test problems using this tool” but i prefer it wouldn’t because it is very tied to `cluster_generator`.

PR #499, opened 2025-12-12
[Bugfix] Fix the use of uninitialized variable in Par_EquilibriumIC

hsinhaoHHuang:
## Issue
The variable `par` in `Par_EquilibriumIC::Set_Velocity()` is uninitialized if `k` loops to `params.Cloud_MassProfNBin-1` without ever `sum_mes>sum_rad` in the for loop.

The resulting velocity of that particle will be polluted and become something unpredictable.
This may lead to extremely large particle velocity and extremely small time-steps.

## Change
Assign a value to `par` when `k==params.Cloud_MassProfNBin-1` in a similar way.

## Note
Related Issue:
- #295 

Related PR:
- #315


hyschive:
@hsinhaoHHuang Thanks for the careful check and the bug fix!

PR #502, opened 2026-01-03
Improve the handling of unphysical results in Hydro_HancockPredict()

hsinhaoHHuang:
## Motivations
- Some numerical instabilities were seen in isolated galaxy simulations with feedback and in hydrodynamical cosmological simulations. The unphysical results in the half-step prediction in the MHM scheme may be the source leading to those artificial explosions.
- The original negative pressure check inside `Hydro_HancockPredict()` was performed on `fcPri[f][4]`, which is the input value before the update. The negative pressure after the update may not be detected, and a floored value is returned at the end.

## Goal
- To improve the accuracy and stability of the half-step Hancock prediction in the MHM scheme.

## Changes
- Bring back the check of negative pressure after the half-step update.
- Add a velocity check against dh/dt in the unphysical checks.
- Add a three-stage rescue iteration locally when the unphysical results are found.
  - First, recalculate the half-step update with more substeps and with the midpoint integration method.
    - Ideally, it should help and could maintain the desired slope. In practice, however, this is not helpful to rescue the unphysical results most of the time. Therefore, in the default setup, `MHM_REPREDICT_ITER_NUM = 2`, this will NOT be used.
  - Second, reduce the slope for data reconstruction and then recalculate the half-step update
    - Although it is more diffuse, this usually helps and is not as extreme as the zero-slope method in the third stage.
  - Third, reset to the cell-centered variables (i.e., zero slope for the half-step update).
    - This is the original method to deal with the unphysical results, and it is the safest.

## Verification Tests
1. `Hydro/Riemann` with `--flu_scheme=MHM`
   - `Riemann_Prob = [0, 1, 2, 3, 4, 5, 9]` was run by **ALWAYS** triggering the rescuing methods (separately) regardless of whether the original update is unphysical to test the correctness of the repredict solver. With the default parameters for the rescue methods, the results can still match the analytical solutions, and there is not much difference from the original update.
     - The results by running with more substeps look almost identical to the original update; while running by the reduced slope, small differences can be seen at some discontinuities.
   - Only `Riemann_Prob = 9` can produce negative pressure in the default half-step update. The 3-stage rescue methods are triggered for that, and the final results can match the analytical solution as usual.
     - Running by more substeps actually looks slightly worse, while running by reduced slope looks better than the default.
2. Isolated galaxy with SN feedback
   The previous explosions are not found anymore with the new method.
3. Cosmological hydrodynamical simulations
   The previous explosions are gone after the negative pressure is properly checked and a lower slope is applied.

## Future Work
- Support `SRHD`
- Turn the hard-coded macros `MHM_REPREDICT_*` into runtime options

## Notes
- The default parameters of safety factors and the number of substeps are just rule-of-thumb estimations. They are not theoretically based or empirically optimized.
- The rescue methods are kind of experimental. Any suggestions on improving the accuracy, stability, and performance (especially for GPUs) for this PR will be greatly appreciated.


src/Auxiliary/Aux_Check_Parameter.cpp
```diff
@@ -1107,6 +1107,31 @@ void Aux_Check_Parameter()
...
+// ------------------------------
+#  if ( FLU_SCHEME == MHM )
+
+#  ifdef MHM_CHECK_PREDICT
+#  ifndef SRHD
+
+   if ( MHM_REPREDICT_ITER_NUM < 1 )
+      Aux_Error( ERROR_INFO, "MHM_REPREDICT_ITER_NUM should be >= 1 for MHM_CHECK_PREDICT !!\n" );
+
+   if ( MHM_REPREDICT_STEPS_SAFE_FAC <= 0 )
```
hyschive, line 1120:
```suggestion
   if ( MHM_REPREDICT_STEPS_SAFE_FAC <= (real)0.0 )
```

src/Model_Hydro/CPU_Hydro/CPU_Shared_DataReconstruction.cpp
```diff
@@ -2161,6 +2217,370 @@ void Hydro_HancockPredict( real fcCon[][NCOMP_LR], const real fcPri[][NCOMP_LR],
...
+// Iter = [       0,          1, ..., IterHalf-1] -> more accurate method
+// Iter = [IterHalf, IterHalf+1, ..., IterNum -2] -> more diffusive method
+   const int IterHalf = (IterNum-1)/2;
+
+// for the first half, try using higher-order integration method and smaller time-step
+   if ( Iter < IterHalf )
+   {
+//    estimate the required numer of sub-steps according to the maximum depletion fraction
+      const real safety_factor    = MHM_REPREDICT_STEPS_SAFE_FAC/(1<<Iter); // e.g., 0.4 -> 0.2 -> 0.1
+      const int  num_substeps_est = (int)ceilf( DeplFracMax/safety_factor );
```
hyschive, line 2428: How about defining `CEIL` in `Macro.h`, similar to `FLOOR`?

src/Auxiliary/Aux_Check_Parameter.cpp
```diff
@@ -1107,6 +1107,31 @@ void Aux_Check_Parameter()
...
+      Aux_Error( ERROR_INFO, "MHM_REPREDICT_STEPS_SAFE_FAC should be > 0 for MHM_CHECK_PREDICT !!\n" );
+
+   if ( MHM_REPREDICT_SUBSTEPS_MAX < 1 )
+      Aux_Error( ERROR_INFO, "MHM_REPREDICT_SUBSTEPS_MAX should be >= 1 for MHM_CHECK_PREDICT !!\n" );
+
+   if ( MHM_REPREDICT_SLOPE_SAFE_FAC <= 0  ||  MHM_REPREDICT_SLOPE_SAFE_FAC >= 1 )
```
hyschive, line 1126:
```suggestion
   if ( MHM_REPREDICT_SLOPE_SAFE_FAC <= (real)0.0  ||  MHM_REPREDICT_SLOPE_SAFE_FAC >= (real)1.0 )
```

src/Auxiliary/Aux_Check_Parameter.cpp
```diff
@@ -1107,6 +1107,31 @@ void Aux_Check_Parameter()
...
 
+// check for MHM
+// ------------------------------
+#  if ( FLU_SCHEME == MHM )
+
+#  ifdef MHM_CHECK_PREDICT
+#  ifndef SRHD
+
+   if ( MHM_REPREDICT_ITER_NUM < 1 )
+      Aux_Error( ERROR_INFO, "MHM_REPREDICT_ITER_NUM should be >= 1 for MHM_CHECK_PREDICT !!\n" );
```
hyschive, line 1118: Always print the incorrect value to facilitate debugging. Update other error messages too.
```suggestion
   if ( MHM_REPREDICT_ITER_NUM < 1 )
      Aux_Error( ERROR_INFO, "MHM_REPREDICT_ITER_NUM (%d) should be >= 1 for MHM_CHECK_PREDICT !!\n", MHM_REPREDICT_ITER_NUM );
```

src/Model_Hydro/CPU_Hydro/CPU_Shared_DataReconstruction.cpp
```diff
@@ -81,6 +81,34 @@ static void Hydro_HancockPredict( real fcCon[][NCOMP_LR], const real fcPri[][NCO
                                   const int NGhost, const int NEle,
                                   const real MinDens, const real MinPres, const real MinEint,
                                   const EoS_t *EoS, const long PassiveFloor );
+#ifdef MHM_CHECK_PREDICT
+GPU_DEVICE
+static bool Hydro_HancockPredict_CheckUnphysical( const real fcCon[][NCOMP_LR], const real dt_dh,
+                                                  const EoS_t *EoS, const long PassiveFloor );
+# ifndef SRHD
```
hyschive, line 88:
```suggestion
#ifndef SRHD
```

src/Model_Hydro/CPU_Hydro/CPU_Shared_DataReconstruction.cpp
```diff
@@ -2099,45 +2135,65 @@ void Hydro_HancockPredict( real fcCon[][NCOMP_LR], const real fcPri[][NCOMP_LR],
 #  endif
 
 
-// check negative, inf, and nan in density, energy, and pressure
+// check unphysical results in the prediction, repredict if needed, and apply floors
 #  ifdef MHM_CHECK_PREDICT
-   bool reset_cell = false;
+   bool repredict = false;           // whether the reprediction is needed
+   real depl_frac_max = MAX_ERROR;   // maximum of the depletion fraction to be found; initialized with a small positive value for safety
```
hyschive, line 2141:
```suggestion
   real depl_frac_max = MAX_ERROR;   // maximum depletion fraction to be determined; initialized with a small positive value for safety
```

hyschive, line 2141: It seems that the initial value here can be arbitrary, as it is always reset by `Hydro_HancockPredict_GetDeplFracMax()`.

src/Model_Hydro/CPU_Hydro/CPU_Shared_DataReconstruction.cpp
```diff
@@ -2161,6 +2217,370 @@ void Hydro_HancockPredict( real fcCon[][NCOMP_LR], const real fcPri[][NCOMP_LR],
 
 
 
+# ifdef MHM_CHECK_PREDICT
+//-------------------------------------------------------------------------------------------------------
+// Function    :  Hydro_HancockPredict_CheckUnphysical
+// Description :  Check the unphysical results after the half time-step update in HancockPredict
```
hyschive, line 2223:
```suggestion
// Description :  Check the unphysical results after the half time-step update in Hydro_HancockPredict()
```

src/Model_Hydro/CPU_Hydro/CPU_Shared_DataReconstruction.cpp
```diff
@@ -2099,45 +2135,65 @@ void Hydro_HancockPredict( real fcCon[][NCOMP_LR], const real fcPri[][NCOMP_LR],
...
    {
-#     ifdef SRHD
-      if (  Hydro_IsUnphysical( UNPHY_MODE_CONS, fcCon[f], NULL_REAL,
-                                EoS->DensEint2Pres_FuncPtr,
-                                EoS->GuessHTilde_FuncPtr, EoS->HTilde2Temp_FuncPtr,
-                                EoS->AuxArrayDevPtr_Flt, EoS->AuxArrayDevPtr_Int, EoS->Table,
-                                PassiveFloor, ERROR_INFO, UNPHY_SILENCE )  )
-         reset_cell = true;
+//    check whether there are unphysical results after the update
+      repredict = Hydro_HancockPredict_CheckUnphysical( fcCon, dt_dh2*2, EoS, PassiveFloor );
```
hyschive, line 2159:
```suggestion
      repredict = Hydro_HancockPredict_CheckUnphysical( fcCon, dt_dh2*(real)2.0, EoS, PassiveFloor );
```

src/Model_Hydro/CPU_Hydro/CPU_Shared_DataReconstruction.cpp
```diff
@@ -2099,45 +2135,65 @@ void Hydro_HancockPredict( real fcCon[][NCOMP_LR], const real fcPri[][NCOMP_LR],
...
+//    break immediately when there is no need to repredict
+      if ( !repredict )   break;
 
-//    set to the cell-centered values before update
-      if ( reset_cell )
+#     ifndef SRHD
+//    repredict with more accurate or safer methods to avoid overshoot
+      if ( repredict_iter != repredict_iter_num-1 )
+      {
+//       decide the maxium depletion fraction of density, energy, and internal energy
```
hyschive, line 2168:
```suggestion
//       determine the maxium depletion fraction of density, energy, and internal energy
```

src/Model_Hydro/CPU_Hydro/CPU_Shared_DataReconstruction.cpp
hyschive, line 2141: For clarity, how about moving this declaration inside `if ( repredict_iter != repredict_iter_num-1 ) { ... }`?

src/Model_Hydro/CPU_Hydro/CPU_Shared_DataReconstruction.cpp
```diff
@@ -2099,45 +2135,65 @@ void Hydro_HancockPredict( real fcCon[][NCOMP_LR], const real fcPri[][NCOMP_LR],
 #  endif
 
 
-// check negative, inf, and nan in density, energy, and pressure
+// check unphysical results in the prediction, repredict if needed, and apply floors
 #  ifdef MHM_CHECK_PREDICT
-   bool reset_cell = false;
+   bool repredict = false;           // whether the reprediction is needed
```
hyschive, line 2140: For clarity, how about moving this declaring inside `for (int repredict_iter=0; repredict_iter<repredict_iter_num; repredict_iter++) { ... }` and changing it to `const bool repredict = Hydro_HancockPredict_CheckUnphysical( ... )`?

src/Model_Hydro/CPU_Hydro/CPU_Shared_DataReconstruction.cpp
```diff
@@ -2099,45 +2135,65 @@ void Hydro_HancockPredict( real fcCon[][NCOMP_LR], const real fcPri[][NCOMP_LR],
...
+         if ( repredict_iter == 0 )
+            depl_frac_max = Hydro_HancockPredict_GetDeplFracMax( fcCon, fcCon_init, fcPri, MinEint, PassiveFloor );
+
+//       invoke the rescue methods to repredict
+         Hydro_HancockPredict_IterateReprediction( fcCon, fcCon_init, dt_dh2,
+                                                   g_cc_array, cc_idx, MinPres, EoS, PassiveFloor,
+                                                   repredict_iter, repredict_iter_num, depl_frac_max );
+
+//       reset the flag for the next check
+         repredict = false;
```
hyschive, line 2179: This reset seems unnecessary.

src/Model_Hydro/CPU_Hydro/CPU_Shared_DataReconstruction.cpp
```diff
@@ -2161,6 +2217,370 @@ void Hydro_HancockPredict( real fcCon[][NCOMP_LR], const real fcPri[][NCOMP_LR],
...
 
+# ifdef MHM_CHECK_PREDICT
+//-------------------------------------------------------------------------------------------------------
+// Function    :  Hydro_HancockPredict_CheckUnphysical
+// Description :  Check the unphysical results after the half time-step update in HancockPredict
+//
+// Note        :  1. Work for the MHM scheme
+//                2. Invoked by Hydro_HancockPredict() only
+//                3. Input variables must be conserved variables
+//                4. Check negative, inf, and nan in density, energy, and pressure and high velocity
```
hyschive, line 2228:
```suggestion
//                4. Check negative, inf, and nan in density, energy, and pressure, as well as excessively high velocity
```

src/Model_Hydro/CPU_Hydro/CPU_Shared_DataReconstruction.cpp
```diff
@@ -2161,6 +2217,370 @@ void Hydro_HancockPredict( real fcCon[][NCOMP_LR], const real fcPri[][NCOMP_LR],
...
+//                5. It stops checking and returns once any unphysical result is found
+//
+// Parameter   :  fcCon             : Face-centered conserved variables to be checked
+//                dt_dh             : Time interval to advance solution / Cell size
+//                EoS               : EoS object
+//                PassiveFloor      : Bitwise flag to specify the passive scalars to be floored
```
hyschive, line 2234:
```suggestion
// Parameter   :  fcCon        : Face-centered conserved variables to be checked
//                dt_dh        : Time interval to advance solution / cell size
//                EoS          : EoS object
//                PassiveFloor : Bitwise flag to specify the passive scalars to be floored
```

src/Model_Hydro/CPU_Hydro/CPU_Shared_DataReconstruction.cpp
```diff
@@ -2161,6 +2217,370 @@ void Hydro_HancockPredict( real fcCon[][NCOMP_LR], const real fcPri[][NCOMP_LR],
...
+//-------------------------------------------------------------------------------------------------------
```
hyschive, line 2235:
```suggestion
// Return      : isUnphysical
//-------------------------------------------------------------------------------------------------------
```

src/Model_Hydro/CPU_Hydro/CPU_Shared_DataReconstruction.cpp
```diff
@@ -2161,6 +2217,370 @@ void Hydro_HancockPredict( real fcCon[][NCOMP_LR], const real fcPri[][NCOMP_LR],
...
+#     else
+//    1. check the value of density
+      if (  Hydro_IsUnphysical_Single( fcCon[f][DENS], "density",  TINY_NUMBER, HUGE_NUMBER, ERROR_INFO, UNPHY_SILENCE )  )
+      {
+         isUnphysical = true;
+         break;
+      }
+
+//    2. check the value of velocity
+      if ( MAX( FABS(fcCon[f][MOMX]), MAX( FABS(fcCon[f][MOMY]), FABS(fcCon[f][MOMZ]) ) )*dt_dh > FABS(fcCon[f][DENS]) )
```
hyschive, line 2266:
```suggestion
      if ( FMAX( FABS(fcCon[f][MOMX]), FMAX( FABS(fcCon[f][MOMY]), FABS(fcCon[f][MOMZ]) ) )*dt_dh > FABS(fcCon[f][DENS]) )
```

src/Model_Hydro/CPU_Hydro/CPU_Shared_DataReconstruction.cpp
```diff
@@ -2161,6 +2217,370 @@ void Hydro_HancockPredict( real fcCon[][NCOMP_LR], const real fcPri[][NCOMP_LR],
...
+
+#     else
+//    1. check the value of density
+      if (  Hydro_IsUnphysical_Single( fcCon[f][DENS], "density",  TINY_NUMBER, HUGE_NUMBER, ERROR_INFO, UNPHY_SILENCE )  )
+      {
+         isUnphysical = true;
+         break;
+      }
+
+//    2. check the value of velocity
```
hyschive, line 2265:
```suggestion
//    2. check the value of velocity: require max(v)*dt < dh
```

src/Model_Hydro/CPU_Hydro/CPU_Shared_DataReconstruction.cpp
```diff
@@ -2161,6 +2217,370 @@ void Hydro_HancockPredict( real fcCon[][NCOMP_LR], const real fcPri[][NCOMP_LR],
...
+//                3. The value will be used to estimate the required number of steps and the reduced slope
+//                   to rescue the unphysical results
+//
+// Parameter   :  fcCon             : Updated face-centered conserved variables
+//                fcCon_init        : Input face-centered conserved variables before update
+//                fcPri_init        : Input face-centered primitive variables before update
+//                MinEint           : Internal energy floors
+//                PassiveFloor      : Bitwise flag to specify the passive scalars to be floored
+//
+// Return      :  depl_frac_max     : Maximum depletion fraction
```
hyschive, line 2323:
```suggestion
// Parameter   :  fcCon         : Updated face-centered conserved variables
//                fcCon_init    : Input face-centered conserved variables before update
//                fcPri_init    : Input face-centered primitive variables before update
//                MinEint       : Internal energy floor
//                PassiveFloor  : Bitwise flag to specify the passive scalars to be floored
//
// Return      :  depl_frac_max : Maximum depletion fraction
```

src/Model_Hydro/CPU_Hydro/CPU_Shared_DataReconstruction.cpp
```diff
@@ -2161,6 +2217,370 @@ void Hydro_HancockPredict( real fcCon[][NCOMP_LR], const real fcPri[][NCOMP_LR],
...
+      Hydro_HancockPredict_RescueByLowerSlopes( fcCon, fcCon_init, dt_dh2, g_cc_array, cc_idx,
+                                                MinPres, EoS, PassiveFloor, reduced_slope_ratio );
+   } // if ( Iter < IterHalf ) ... else ...
+
+}
+
+
+
+//-------------------------------------------------------------------------------------------------------
+// Function    :  Hydro_HancockPredict_RescueByHigherSteps
```
hyschive, line 2453: "HigherSteps" sounds a bit unusual. How about "SubSteps"?

src/Model_Hydro/CPU_Hydro/CPU_Shared_DataReconstruction.cpp
```diff
@@ -2161,6 +2217,370 @@ void Hydro_HancockPredict( real fcCon[][NCOMP_LR], const real fcPri[][NCOMP_LR],
...
+   return isUnphysical;
+
+} // FUNCTION : Hydro_HancockPredict_CheckUnphysical
+
+
+
+# ifndef SRHD
+//-------------------------------------------------------------------------------------------------------
+// Function    :  Hydro_HancockPredict_GetDeplFracMax
+// Description :  Get the maximum of depletion fraction of the face-centered variables after update among all faces
```
hyschive, line 2310:
```suggestion
// Description :  Get the maximum depletion fraction of the face-centered variables after the update over all faces
```

src/Model_Hydro/CPU_Hydro/CPU_Shared_DataReconstruction.cpp
```diff
@@ -2161,6 +2217,370 @@ void Hydro_HancockPredict( real fcCon[][NCOMP_LR], const real fcPri[][NCOMP_LR],
...
+real Hydro_HancockPredict_GetDeplFracMax( const real fcCon[][NCOMP_LR],
+                                          const real fcCon_init[][NCOMP_LR], const real fcPri_init[][NCOMP_LR],
+                                          const real MinEint, const long PassiveFloor )
+{
+
+// initialize it with a small but non-zero value
+   real depl_frac_max = MAX_ERROR;
+
+// depletion fraction is defined as (var_initial - var_updated)/var_initial
+// positive (>0) when the variable decreased after update
```
hyschive, line 2335:
```suggestion
// positive (>0) when the variable decreases after update
```

src/Model_Hydro/CPU_Hydro/CPU_Shared_DataReconstruction.cpp
```diff
@@ -2161,6 +2217,370 @@ void Hydro_HancockPredict( real fcCon[][NCOMP_LR], const real fcPri[][NCOMP_LR],
...
+//           =1 leads to vacuum, >1 is unphysical, and the larger the dangerous
+// negative (<0) when the variable increased after update, and it is safe
```
hyschive, line 2337:
```suggestion
//           =1 leads to vacuum, >1 is unphysical, and larger values are more dangerous
// negative (<0) when the variable increases after update, which is safe
```

src/Model_Hydro/CPU_Hydro/CPU_Shared_DataReconstruction.cpp
```diff
@@ -2161,6 +2217,370 @@ void Hydro_HancockPredict( real fcCon[][NCOMP_LR], const real fcPri[][NCOMP_LR],
...
+   {
+//    reduce the slope according to the maximum depletion fraction
+      const real safety_factor       = MHM_REPREDICT_SLOPE_SAFE_FAC/(1<<(Iter-IterHalf)); // e.g., 0.9 -> 0.45 -> 0.225
+      const real reduced_slope_ratio = MIN( safety_factor/DeplFracMax, safety_factor );
+
+      Hydro_HancockPredict_RescueByLowerSlopes( fcCon, fcCon_init, dt_dh2, g_cc_array, cc_idx,
+                                                MinPres, EoS, PassiveFloor, reduced_slope_ratio );
+   } // if ( Iter < IterHalf ) ... else ...
+
+}
```
hyschive, line 2448:
```suggestion
} // FUNCTION :  Hydro_HancockPredict_IterateReprediction
```

src/Model_Hydro/CPU_Hydro/CPU_Shared_DataReconstruction.cpp
```diff
@@ -2161,6 +2217,370 @@ void Hydro_HancockPredict( real fcCon[][NCOMP_LR], const real fcPri[][NCOMP_LR],
...
+// negative (<0) when the variable increased after update, and it is safe
+
+// loop through all the faces to find the maximum
+   for (int f=0; f<6; f++)
+   {
+//    density depletion fraction
+      depl_frac_max = MAX( (real)1.0-fcCon[f][DENS]/fcCon_init[f][DENS], depl_frac_max );
+
+//    energy depletion fraction
+      depl_frac_max = MAX( (real)1.0-fcCon[f][ENGY]/fcCon_init[f][ENGY], depl_frac_max );
```
hyschive, line 2346:
```suggestion
      depl_frac_max = FMAX( (real)1.0-fcCon[f][DENS]/fcCon_init[f][DENS], depl_frac_max );

//    energy depletion fraction
      depl_frac_max = FMAX( (real)1.0-fcCon[f][ENGY]/fcCon_init[f][ENGY], depl_frac_max );
```

hyschive, line 2346: Should this be applied only when `BAROTROPIC_EOS` is defined?

src/Model_Hydro/CPU_Hydro/CPU_Shared_DataReconstruction.cpp
```diff
@@ -2161,6 +2217,370 @@ void Hydro_HancockPredict( real fcCon[][NCOMP_LR], const real fcPri[][NCOMP_LR],
...
+
+//    internal energy depletion fraction
+#     ifndef BAROTROPIC_EOS
+//    compute the magnetic energy
+#     ifdef MHD
+      const real Emag      = (real)0.5*( SQR(     fcCon[f][MAG_OFFSET+0]) + SQR(     fcCon[f][MAG_OFFSET+1]) + SQR(     fcCon[f][MAG_OFFSET+2]) );
+      const real initlEmag = (real)0.5*( SQR(fcCon_init[f][MAG_OFFSET+0]) + SQR(fcCon_init[f][MAG_OFFSET+1]) + SQR(fcCon_init[f][MAG_OFFSET+2]) );
```
hyschive, line 2353: How about renaming it to `initEmag` for consistency?

src/Model_Hydro/CPU_Hydro/CPU_Shared_DataReconstruction.cpp
```diff
@@ -2161,6 +2217,370 @@ void Hydro_HancockPredict( real fcCon[][NCOMP_LR], const real fcPri[][NCOMP_LR],
...
+#     endif // MHD
+
+//    compute the initial internal energy; check minimum in case it is negative in the input
+      const bool CheckMinEint_Yes = true;
+      const real initlEint = Hydro_Con2Eint( fcCon_init[f][DENS], fcCon_init[f][MOMX],
+                                             fcCon_init[f][MOMY], fcCon_init[f][MOMZ], fcCon_init[f][ENGY],
+                                             CheckMinEint_Yes, MinEint, PassiveFloor, initlEmag,
+                                             NULL, NULL, NULL, NULL, NULL );
+
+//    estimate the decreased amount of internal energy from the initial change rate with linearization
```
hyschive, line 2365: Could you clarify why we don't also compute it as "(var_initial - var_updated)/var_initial"?

src/Model_Hydro/CPU_Hydro/CPU_Shared_DataReconstruction.cpp
```diff
@@ -2161,6 +2217,370 @@ void Hydro_HancockPredict( real fcCon[][NCOMP_LR], const real fcPri[][NCOMP_LR],
...
+#     ifdef MHD
+      const real Emag      = (real)0.5*( SQR(     fcCon[f][MAG_OFFSET+0]) + SQR(     fcCon[f][MAG_OFFSET+1]) + SQR(     fcCon[f][MAG_OFFSET+2]) );
+      const real initlEmag = (real)0.5*( SQR(fcCon_init[f][MAG_OFFSET+0]) + SQR(fcCon_init[f][MAG_OFFSET+1]) + SQR(fcCon_init[f][MAG_OFFSET+2]) );
+#     else
+      const real initlEmag = NULL_REAL;
+#     endif // MHD
+
+//    compute the initial internal energy; check minimum in case it is negative in the input
+      const bool CheckMinEint_Yes = true;
+      const real initlEint = Hydro_Con2Eint( fcCon_init[f][DENS], fcCon_init[f][MOMX],
```
hyschive, line 2360: How about renaming it to `initEint` for consistency?

src/Model_Hydro/CPU_Hydro/CPU_Shared_DataReconstruction.cpp
```diff
@@ -2161,6 +2217,370 @@ void Hydro_HancockPredict( real fcCon[][NCOMP_LR], const real fcPri[][NCOMP_LR],
...
+
+   return depl_frac_max;
+
+} // FUNCTION : Hydro_HancockPredict_GetDeplFracMax
+
+
+
+//-------------------------------------------------------------------------------------------------------
+// Function    :  Hydro_HancockPredict_IterateReprediction
+// Description :  Redo the HancockPredict evolution by iterations of different trials
```
hyschive, line 2389:
```suggestion
// Description :  Recompute the Hancock prediction through iterative trials
```

src/Model_Hydro/CPU_Hydro/CPU_Shared_DataReconstruction.cpp
```diff
@@ -2161,6 +2217,370 @@ void Hydro_HancockPredict( real fcCon[][NCOMP_LR], const real fcPri[][NCOMP_LR],
...
+//
+// Note        :  1. Work for the MHM scheme
+//                2. Invoked by Hydro_HancockPredict() only when unphysical result is found
+//                3. Invoke different rescue methods accordingly to the iteration times
```
hyschive, line 2393:
```suggestion
//                3. Invoke different rescue methods depending on the iteration count
```

src/Model_Hydro/CPU_Hydro/CPU_Shared_DataReconstruction.cpp
```diff
@@ -2161,6 +2217,370 @@ void Hydro_HancockPredict( real fcCon[][NCOMP_LR], const real fcPri[][NCOMP_LR],
...
+//                g_cc_array        : Array storing the cell-centered conserved variables for checking
+//                                    negative density and pressure
+//                                    --> It is just the input array Flu_Array_In[]
+//                cc_idx            : Index for accessing g_cc_array[]
+//                MinPres           : Pressure floor
+//                EoS               : EoS object
+//                PassiveFloor      : Bitwise flag to specify the passive scalars to be floored
+//                Iter              : Current iteration
+//                IterNum           : Total number of iterations
+//                DeplFracMax       : Maximum depletion fraction based on the initial prediction
```
hyschive, line 2408:
```suggestion
// Parameter   :  fcCon        : Face-centered conserved variables to be updated
//                fcCon_init   : Input face-centered conserved variables before update
//                dt_dh2       : 0.5 * time interval to advance solution / cell size
//                g_cc_array   : Array storing the cell-centered conserved variables for checking
//                               negative density and pressure
//                               --> It is just the input array Flu_Array_In[]
//                cc_idx       : Index for accessing g_cc_array[]
//                MinPres      : Pressure floor
//                EoS          : EoS object
//                PassiveFloor : Bitwise flag to specify the passive scalars to be floored
//                Iter         : Current iteration
//                IterNum      : Total number of iterations
//                DeplFracMax  : Maximum depletion fraction based on the initial prediction
```

src/Model_Hydro/CPU_Hydro/CPU_Shared_DataReconstruction.cpp
```diff
@@ -2161,6 +2217,370 @@ void Hydro_HancockPredict( real fcCon[][NCOMP_LR], const real fcPri[][NCOMP_LR],
...
+//-------------------------------------------------------------------------------------------------------
+GPU_DEVICE
+void Hydro_HancockPredict_IterateReprediction( real fcCon[][NCOMP_LR],
+                                               const real fcCon_init[][NCOMP_LR], const real dt_dh2,
+                                               const real g_cc_array[][ CUBE(FLU_NXT) ], const int cc_idx,
+                                               const real MinPres, const EoS_t *EoS, const long PassiveFloor,
+                                               const int Iter, const int IterNum, const real DeplFracMax )
+{
+
+// iteration division of rescuing methods
```
hyschive, line 2418:
```suggestion
// select rescue methods based on the iteration count
```

src/Model_Hydro/CPU_Hydro/CPU_Shared_DataReconstruction.cpp
```diff
@@ -2161,6 +2217,370 @@ void Hydro_HancockPredict( real fcCon[][NCOMP_LR], const real fcPri[][NCOMP_LR],
...
+// Iter = [       0,          1, ..., IterHalf-1] -> more accurate method
+// Iter = [IterHalf, IterHalf+1, ..., IterNum -2] -> more diffusive method
+   const int IterHalf = (IterNum-1)/2;
+
+// for the first half, try using higher-order integration method and smaller time-step
```
hyschive, line 2423:
```suggestion
// for the first half, try higher-order integration method and smaller time-step
```

src/Model_Hydro/CPU_Hydro/CPU_Shared_DataReconstruction.cpp
```diff
@@ -2161,6 +2217,370 @@ void Hydro_HancockPredict( real fcCon[][NCOMP_LR], const real fcPri[][NCOMP_LR],
...
+   if ( Iter < IterHalf )
+   {
+//    estimate the required numer of sub-steps according to the maximum depletion fraction
+      const real safety_factor    = MHM_REPREDICT_STEPS_SAFE_FAC/(1<<Iter); // e.g., 0.4 -> 0.2 -> 0.1
+      const int  num_substeps_est = (int)ceilf( DeplFracMax/safety_factor );
+//    or at least increase by iterations
+      const int  num_substeps_min = 1<<Iter;                                // 1 -> 2 -> 4
+      const int  num_substeps_max = MHM_REPREDICT_SUBSTEPS_MAX;             // to avoid too many sub-steps, for the performance consideration
```
hyschive, line 2431:
```suggestion
      const int  num_substeps_max = MHM_REPREDICT_SUBSTEPS_MAX;             // to avoid too many sub-steps for performance considerations
```

src/Model_Hydro/CPU_Hydro/CPU_Shared_DataReconstruction.cpp
```diff
@@ -2161,6 +2217,370 @@ void Hydro_HancockPredict( real fcCon[][NCOMP_LR], const real fcPri[][NCOMP_LR],
...
+
+      Hydro_HancockPredict_RescueByHigherSteps( fcCon, fcCon_init, dt_dh2,
+                                                MinPres, EoS, PassiveFloor, num_substeps );
+   }
+// for the second half, try reducing the slope for data reconstruction
+   else // if ( Iter < IterHalf )
+   {
+//    reduce the slope according to the maximum depletion fraction
+      const real safety_factor       = MHM_REPREDICT_SLOPE_SAFE_FAC/(1<<(Iter-IterHalf)); // e.g., 0.9 -> 0.45 -> 0.225
+      const real reduced_slope_ratio = MIN( safety_factor/DeplFracMax, safety_factor );
```
hyschive, line 2442:
```suggestion
      const real reduced_slope_ratio = FMIN( safety_factor/DeplFracMax, safety_factor );
```

src/Model_Hydro/CPU_Hydro/CPU_Shared_DataReconstruction.cpp
```diff
@@ -2161,6 +2217,370 @@ void Hydro_HancockPredict( real fcCon[][NCOMP_LR], const real fcPri[][NCOMP_LR],
...
+                                                MinPres, EoS, PassiveFloor, reduced_slope_ratio );
+   } // if ( Iter < IterHalf ) ... else ...
+
+}
+
+
+
+//-------------------------------------------------------------------------------------------------------
+// Function    :  Hydro_HancockPredict_RescueByHigherSteps
+// Description :  Evolve the face-centered variables by half time-step by subcyling steps
```
hyschive, line 2454:
```suggestion
// Description :  Evolve the face-centered variables using both half-step prediction and subcycling
```

src/Model_Hydro/CPU_Hydro/CPU_Shared_DataReconstruction.cpp
```diff
@@ -2161,6 +2217,370 @@ void Hydro_HancockPredict( real fcCon[][NCOMP_LR], const real fcPri[][NCOMP_LR],
...
+//
+// Note        :  1. Work for the MHM scheme
+//                2. Invoked by Hydro_HancockPredict_IterateReprediction() when unphysical result is found in Hydro_HancockPredict()
+//                3. Input variables must be conserved variables
+//                4. Warning: This is not helpful for rescuing most of the time and can waste time
```
hyschive, line 2459:
```suggestion
//                4. Warning: This is often ineffective for rescue and may waste time
```

src/Model_Hydro/CPU_Hydro/CPU_Shared_DataReconstruction.cpp
```diff
@@ -2161,6 +2217,370 @@ void Hydro_HancockPredict( real fcCon[][NCOMP_LR], const real fcPri[][NCOMP_LR],
...
+//
+// Parameter   :  fcCon             : Face-centered conserved variables to be updated
+//                fcCon_init        : Input face-centered conserved variables before update
+//                dt_dh2            : 0.5 * Time interval to advance solution / Cell size
+//                MinPres           : Pressure floors
+//                EoS               : EoS object
+//                PassiveFloor      : Bitwise flag to specify the passive scalars to be floored
+//                NumSubSteps       : Number of sub-steps to evolve
```
hyschive, line 2467:
```suggestion
// Parameter   :  fcCon        : Face-centered conserved variables to be updated
//                fcCon_init   : Input face-centered conserved variables before update
//                dt_dh2       : 0.5 * time interval to advance solution / cell size
//                MinPres      : Pressure floor
//                EoS          : EoS object
//                PassiveFloor : Bitwise flag to specify the passive scalars to be floored
//                NumSubSteps  : Number of sub-steps to evolve
```

src/Model_Hydro/CPU_Hydro/CPU_Shared_DataReconstruction.cpp
```diff
@@ -2161,6 +2217,370 @@ void Hydro_HancockPredict( real fcCon[][NCOMP_LR], const real fcPri[][NCOMP_LR],
...
+// restore the initial face-centered variables
+   for (int f=0; f<6; f++)
+   for (int v=0; v<NCOMP_TOTAL; v++)
+      fcCon[f][v] = fcCon_init[f][v];
+
+// adopt the midpoint method for each sub-step of integration
+// dt_sub is the sub-step time-step; evolution destination is 0.5*dt
+   const real dt_sub_dh  = dt_dh2/NumSubSteps;    // for full sub-step update
+   const real dt_sub_dh2 = (real)0.5*dt_sub_dh;   // for half sub-step update
+   for (int substep=0; substep<NumSubSteps; substep++)
```
hyschive, line 2485:
```suggestion
   const real dt_sub_dh  = dt_dh2/NumSubSteps;    // for full sub-step update
   const real dt_sub_dh2 = (real)0.5*dt_sub_dh;   // for half sub-step update
   
   for (int substep=0; substep<NumSubSteps; substep++)
```

src/Model_Hydro/CPU_Hydro/CPU_Shared_DataReconstruction.cpp
```diff
@@ -2161,6 +2217,370 @@ void Hydro_HancockPredict( real fcCon[][NCOMP_LR], const real fcPri[][NCOMP_LR],
...
+// Parameter   :  fcCon             : Face-centered conserved variables to be updated
+//                fcCon_init        : Input face-centered conserved variables before update
+//                dt_dh2            : 0.5 * Time interval to advance solution / Cell size
+//                MinPres           : Pressure floors
+//                EoS               : EoS object
+//                PassiveFloor      : Bitwise flag to specify the passive scalars to be floored
+//                NumSubSteps       : Number of sub-steps to evolve
+//-------------------------------------------------------------------------------------------------------
+GPU_DEVICE
+void Hydro_HancockPredict_RescueByHigherSteps( real fcCon[][NCOMP_LR],
```
hyschive, line 2470: IIUC, this routine does _NOT_ update the B field. In other words, it always uses the original B field stored in `fcCon_init`, which is defined before any evolution takes place. I'm not sure how this would affect the results.

At a minimum, add a comment about this concern.

This is related to the general comment I made in this review.

src/Model_Hydro/CPU_Hydro/CPU_Shared_DataReconstruction.cpp
```diff
@@ -2161,6 +2217,370 @@ void Hydro_HancockPredict( real fcCon[][NCOMP_LR], const real fcPri[][NCOMP_LR],
...
+//                                    negative density and pressure
+//                                    --> It is just the input array Flu_Array_In[]
+//                cc_idx            : Index for accessing g_cc_array[]
+//                MinPres           : Density, pressure, and internal energy floors
+//                EoS               : EoS object
+//                PassiveFloor      : Bitwise flag to specify the passive scalars to be floored
+//                ReducedSlopeRatio : Reduced slope / original slope
+//-------------------------------------------------------------------------------------------------------
+GPU_DEVICE
+void Hydro_HancockPredict_RescueByLowerSlopes( real fcCon[][NCOMP_LR],
```
hyschive, line 2550: IIUC, this routine does _NOT_ change the B field. In other words, it always uses the B field stored in `fcCon_init`, which is computed using the original data reconstruction scheme. I'm not sure whether this minor inconsistency (i.e., adopting different reconstruction methods for the fluid and B fields) would affect the results.

At a minimum, add a comment about this concern.

This is related to the general comment I made in this review.

src/Model_Hydro/CPU_Hydro/CPU_Shared_DataReconstruction.cpp
```diff
@@ -2161,6 +2217,370 @@ void Hydro_HancockPredict( real fcCon[][NCOMP_LR], const real fcPri[][NCOMP_LR],
...
+// Function    :  Hydro_HancockPredict_RescueByLowerSlopes
+// Description :  Evolve the face-centered variables by half time-step with a reduced slope
+//
+// Note        :  1. Work for the MHM scheme
+//                2. Invoked by Hydro_HancockPredict_IterateReprediction() when unphysical result is found in Hydro_HancockPredict()
+//                3. Input variables must be conserved variables
+//
+// Parameter   :  fcCon             : Face-centered conserved variables to be updated
+//                fcCon_init        : Input face-centered conserved variables before update
+//                dt_dh2            : 0.5 * Time interval to advance solution / Cell size
```
hyschive, line 2539:
```suggestion
//                dt_dh2            : 0.5 * time interval to advance solution / cell size
```

src/Model_Hydro/CPU_Hydro/CPU_Shared_DataReconstruction.cpp
```diff
@@ -2161,6 +2217,370 @@ void Hydro_HancockPredict( real fcCon[][NCOMP_LR], const real fcPri[][NCOMP_LR],
...
+//                g_cc_array        : Array storing the cell-centered conserved variables for checking
+//                                    negative density and pressure
+//                                    --> It is just the input array Flu_Array_In[]
+//                cc_idx            : Index for accessing g_cc_array[]
+//                MinPres           : Density, pressure, and internal energy floors
```
hyschive, line 2544:
```suggestion
//                MinPres           : Pressure floor
```

src/Model_Hydro/CPU_Hydro/CPU_Shared_DataReconstruction.cpp
```diff
@@ -2161,6 +2217,370 @@ void Hydro_HancockPredict( real fcCon[][NCOMP_LR], const real fcPri[][NCOMP_LR],
...
+                                               const real fcCon_init[][NCOMP_LR], const real dt_dh2,
+                                               const real g_cc_array[][ CUBE(FLU_NXT) ], const int cc_idx,
+                                               const real MinPres, const EoS_t *EoS, const long PassiveFloor,
+                                               const real ReducedSlopeRatio )
+{
+
+// reconstruct the face-centered variables with the reduced slope
+   for (int f=0; f<6; f++)
+   for (int v=0; v<NCOMP_TOTAL; v++)
+      fcCon[f][v] = g_cc_array[v][cc_idx] + ReducedSlopeRatio * ( fcCon_init[f][v] - g_cc_array[v][cc_idx] );
```
hyschive, line 2560: Add a comment somewhere in this routine clarifying that this rescue approach assumes a piecewise linear distribution even when adopting the PPM data reconstruction.

include/CUFLU.h
```diff
@@ -261,6 +261,14 @@
 // check unphysical results in the MHM half-step prediction
 #if ( FLU_SCHEME == MHM )
 #  define MHM_CHECK_PREDICT
+#ifdef MHM_CHECK_PREDICT
+#ifndef SRHD
+#  define MHM_REPREDICT_ITER_NUM                2
+#  define MHM_REPREDICT_STEPS_SAFE_FAC  (real)0.4
+#  define MHM_REPREDICT_SLOPE_SAFE_FAC  (real)0.9
+#  define MHM_REPREDICT_SUBSTEPS_MAX            4
```
hyschive, line 269: Clarify the usage and typical ranges of these parameters.

hyschive (commented):
@hsinhaoHHuang This PR looks good. Thanks for the great effort!

One concern is that I’m not sure whether this prediction works properly for MHD. Check the related comments in `Hydro_HancockPredict_RescueByLowerSlopes()` and `Hydro_HancockPredict_RescueByHigherSteps()`.

src/Model_Hydro/CPU_Hydro/CPU_Shared_DataReconstruction.cpp
hsinhaoHHuang: Done.

src/Model_Hydro/CPU_Hydro/CPU_Shared_DataReconstruction.cpp
hsinhaoHHuang: Removed.

src/Model_Hydro/CPU_Hydro/CPU_Shared_DataReconstruction.cpp
hsinhaoHHuang: Done.

src/Model_Hydro/CPU_Hydro/CPU_Shared_DataReconstruction.cpp
hsinhaoHHuang: Done.

src/Model_Hydro/CPU_Hydro/CPU_Shared_DataReconstruction.cpp
hsinhaoHHuang: Done.

src/Model_Hydro/CPU_Hydro/CPU_Shared_DataReconstruction.cpp
hsinhaoHHuang: Done.

src/Model_Hydro/CPU_Hydro/CPU_Shared_DataReconstruction.cpp
hsinhaoHHuang: Yes, it is. Would it be better not to set an initial value here?

hsinhaoHHuang: Currently, I declare `depl_frac_max` outside the `for (int repredict_iter=0; repredict_iter<repredict_iter_num; repredict_iter++)` loop because it is only computed once when `repredict_iter == 0 ', but it will be used across iterations. Moving the declaration inside may require computing it on every iteration.

src/Model_Hydro/CPU_Hydro/CPU_Shared_DataReconstruction.cpp
hsinhaoHHuang: Do you mean when `BAROTROPIC_EOS` is undefined?

I added it here simply because I thought it was a quantity that should remain positive. I can put it inside `#ifndef BAROTROPIC_EOS` to be consistent with the energy check in `Hydro_HancockPredict_CheckUnphysical()`.

include/CUFLU.h
hsinhaoHHuang: Added.

src/Model_Hydro/CPU_Hydro/CPU_Shared_DataReconstruction.cpp
hsinhaoHHuang: My concern is that, unlike density and total energy, which can be updated directly from their initial fluxes, the change in internal energy is more indirect. If the updated density or energy is already wrong (e.g., a negative density), the computed internal energy may be unrealistic, and the estimation here may be incorrect and may not be useful.

Does this make sense to you?

src/Model_Hydro/CPU_Hydro/CPU_Shared_DataReconstruction.cpp
hsinhaoHHuang: I'm not familiar with the data reconstruction process.
But my intention here is actually to preserve the original shape of the distribution as much as possible.
If we see the distribution inside the cell as a parabola
$a_j^I(x) = a_{j,-} + \xi [ \Delta a_j + a_{6,j} (1-\xi)]$,
where $0\le \xi \equiv \frac{x-jh}{h}\le 1$, $\Delta a_j\equiv(a_{j,+}-a_{j,-})$, and $a_{6,j}\equiv 6[\langle a\rangle_j^n-\frac{1}{2}(a_{j,-}+a_{j,+})]$.
(Eq. (7), Colella & Sekora 2008)
This can be rewritten as
$a_j^I(x) = \langle a\rangle_j^n + (\xi-\frac{1}{2})\Delta a_j + a_{6,j}[\xi(1-\xi)-\frac{1}{6}]$.
By multiplying a common factor to $a(x_{\rm face})-\langle a\rangle_j^n$ on all the faces here, the distribution can still be seen as a flatter parabola (smaller slope and curvature):
$a_j^{I,{\rm new}}(x) = \langle a\rangle_j^n + (\xi-\frac{1}{2})\Delta a_j^{\rm reduced} + a_{6,j}^{\rm reduced}[\xi(1-\xi)-\frac{1}{6}]$.
At least, the slopes from the center to the two faces, `(fc_R - cc)` and `(cc - fc_L)`, are still different, and their relative ratio is maintained. In my understanding, this is not piecewise linear.

The fix here cannot yield the same results as redoing the entire data reconstruction process, since we are working with conserved variables and don't have neighbors' information. So, I guess it is not PPM strictly.
Do you have any suggestions for this?

src/Model_Hydro/CPU_Hydro/CPU_Shared_DataReconstruction.cpp
hsinhaoHHuang: Yes, it uses the original B field.

In this routine, I treat the half-step evolution as an initial value problem with coupled equations. This may differ from the actual solution considering its neighboring cells. But at least all of the variables are in place.
From what I understand, the B field is updated by the initial edge-centered EMF in `MHD_UpdateMagnetic_Half`
However, it does not appear to be trivial to update the EMF.

Aside from a comment, would you recommend excluding MHD from the reprediction, similarly to `SRHD`?

src/Model_Hydro/CPU_Hydro/CPU_Shared_DataReconstruction.cpp
hsinhaoHHuang: I think it is possible to compute the cell-centered B field first, do the same reconstruction with a reduced slope for the transverse B field as the other field here, and then call `MHD_UpdateMagnetic_Half()` again.

But I am wondering: it seems this inconsistency also exists in the original version, where we ignore the B field when we decide to reset the cell to the cell-centered values [here](https://github.com/gamer-project/gamer/blob/main/src/Model_Hydro/CPU_Hydro/CPU_Shared_DataReconstruction.cpp#L2135). Is there a reason for this?

hsinhaoHHuang (commented):
@hyschive Thank you for your review.
I updated this PR to address some of the comments and left some points for discussion, on which your feedback will be greatly appreciated. Please take a look whenever you have time. Thank you so much.

PR #504, opened 2026-01-03
Skip dual-energy fix for unphysical results in Hydro_FullStepUpdate()

hsinhaoHHuang:
## Motivations
In this PR, I propose we skip the dual-energy fix inside `Hydro_FullStepUpdate` when the results are unphysical.

If there are some reasons that the Riemann solver gets large fluxes, the full-step updated density can be low, the velocity can be high, the kinetic energy may be unphysically large, the total energy can be negative or positive, but the internal energy can be largely negative.

We expect the dual-energy fix to work when the internal energy is low or even slightly negative while other variables are normal. But when the other variables are obviously unphysical, we should not update the total energy by dual-energy fix and let the unphysical results be detected by the check later to invoke 1st-order flux correction or `AUTO_REDUCE_DT`.

The dual-energy fix may falsely mask the unphysical results updated by the full-step fluxes.

## Example
This is one real case in my isolated galaxy simulations:
Input
`[Dens, MomX, MomY, MomZ, Engy] = [ 1.67E-06, -1.67E-04, 1.59E-03, -1.18E-03, 5.07E+01 ]`
so
`[DENS, VELX, VELY, VELZ, EINT] = [ 1.67E-06, -9.99E+01, 9.57E+02, -7.08E+02, 4.96E+01 ]`

After the full-step update, it becomes
Output
`[Dens, MomX, MomY, MomZ, Engy] = [ 4.52E-07, -2.52E-03, -3.98E-02, -4.54E-04, -5.16E+01 ]`
which is 
`[DENS, VELX, VELY, VELZ, EINT] = [ 4.52E-07, -5.58E+03, -8.80E+04, -1.00E+03, -1.81E+03 ]`

The total energy is negative, and the velocity is unreasonably large, so the internal energy is largely negative.

After the dual-energy fix inside `Hydro_FullStepUpdate()`, the total energy will be updated by the sum of the kinetic energy and the internal energy from dual energy and become largely positive (dominated by the kinetic energy).
Output `Engy` after dual-energy fix = `1.7622859e+03`, and the internal energy is also positive.

In `Flu_Close()->CorrectUnphysical()`, we will not notice there are unphysical problems in this cell.
The large total energy will cause other problems later, such as an explosion.

## Changes
In `Hydro_FullStepUpdate()`, only apply the dual-energy fix when the density is positive and the kinetic energy is less than twice the total energy. In this way, the negative total energy or negative internal energy will be kept and noticed.


src/Model_Hydro/CPU_Hydro/CPU_Shared_FullStepUpdate.cpp
```diff
@@ -160,6 +160,12 @@ void Hydro_FullStepUpdate( const real g_Input[][ CUBE(FLU_NXT) ], real g_Output[
 //    we no longer apply density and pressure floors here since we want to enable 1st-order-flux correction for that
 //    Output_1Cell[DENS] = FMAX( Output_1Cell[DENS], MinDens );
 
+//    we do not apply the dual-energy fix to the energy
+//    when the density is negative or Ekin >= 2*Etot (could be caused by an extremely low density)
+//    --> such that the unphysical energy can be caught by the checks in Flu_Close()->CorrectUnphysical()
```
hyschive, line 165: It would be helpful to further clarify the following points:
- Why would an extremely low density cause Ekin >= 2*Etot? Consider giving a concrete example.
- When "density is negative or Ekin >= 2*Etot", what would happen if the dual-energy fix were still applied? My understanding is that the internal energy becomes reasonable, but the total energy becomes unphysically large due to the extremely large kinetic energy (resulting from very low density and finite momentum). Consider providing a concrete example for this case as well.

hsinhaoHHuang: I made up an example to explain it. Would you think it is helpful? Or is it too long? Please let me know if there is still anything unclear. And any comment is appreciated. Thank you!

```suggestion
//    when the density is negative or Ekin >= 2*Etot (could be caused by an extremely low density)
//    --> such that the unphysical energy can be caught by the checks in Flu_Close()->CorrectUnphysical()
//
//    note that ( i) "Etot*2 > Ekin" condition also implies "Etot > 0"
//              (ii) "Ekin >= 2*Etot" means "Eint <= -Etot" and "Eint <= -1/2*Ekin",
//                    in other words, Eint has a large negative value compared to Etot and Ekin
//
//    for example, assume the typical input fluid before the hydro solver has
//          (Dens = 1.00, MomX = 1.00, MomY = 1.00, MomZ = 1.00, Etot = 2.00) --> (Ekin = 1.50, Eint = 0.50):
//
//    (a) it is okay to apply the dual-energy fix when the output fluid has
//          (Dens = 0.10, MomX = 0.10, MomY = 0.10, MomZ = 0.10, Etot = 0.16) --> (Ekin = 0.15, Eint = 0.01),
//        where Eint/Ekin = 1./15. is assumed to be lower than DualEnergySwitch
//
//    (b) it is also okay to apply the dual-energy fix when the output fluid has a slightly negative Eint, such as
//          (Dens = 0.10, MomX = 0.10, MomY = 0.10, MomZ = 0.10, Etot = 0.12) --> (Ekin = 0.15, Eint = -0.03)
//
//    in these cases, the Etot will be fixed to physically correct Etot' = Ekin + (internal energy from DUAL)
//
//    however,
//    (c) if the output density is low, the fluid marginally has
//          (Dens = 0.05, MomX = 0.10, MomY = 0.10, MomZ = 0.10, Etot = 0.15) --> (Ekin = 0.30, Eint = -0.15).
//        in this case, the dual-energy fixed Etot' = Ekin + (internal energy from DUAL) will become at least 2x
//
//    (d) it is worse if the output fluid has an even lower density while momentum remains finite,
//          (Dens = 1e-3, MomX = 0.10, MomY = 0.10, MomZ = 0.10, Etot = 0.15) --> (Ekin = 15.0, Eint = -14.85),
//        the dual-energy fixed Etot' will be unphysically large due to the large kinetic energy
//        in this case, Etot' and Eint' are all positive and may pass the unphysical check
//
//    in these cases, we should leave the negative Eint to be handled by Flu_Close()
```

hyschive: > note that ( i) "Etot*2 > Ekin" condition also implies "Etot > 0"

I'm confused by this comment. I thought you should consider the case "Ekin >= 2*Etot", for which Etot can be posive or negative.

> (c) if the output density is low, the fluid marginally has

Change to "(c) if the output density is low, for example,"

> the dual-energy fixed

Change to "the dual-energy-fixed Etot'"

hsinhaoHHuang: Etot can be positive or negative, and "Ekin >= 2*Etot" works in either case.

What I want to point out in this comment is that this check also ensures that the dual-energy fix is only applied when "Etot > 0", which is a less stringent but more basic condition.
If Etot <= 0, the condition "Ekin >= 2*Etot" is always true, so we do not apply the dual-energy fix.

I have updated the comments:
```suggestion
//    when the density is negative or Ekin >= 2*Etot (could be caused by an extremely low density)
//    --> such that the unphysical energy can be caught by the checks in Flu_Close()->CorrectUnphysical()
//
//    note that ( i) Passing "Etot*2 > Ekin" for the dual-energy fix implies "Etot > 0"
//              (ii) "Ekin >= 2*Etot" means "Eint <= -Etot" and "Eint <= -1/2*Ekin",
//                    in other words, Eint has a large negative value compared to Etot and Ekin
//
//    for example, assume the typical input fluid before the hydro solver has
//          (Dens = 1.00, MomX = 1.00, MomY = 1.00, MomZ = 1.00, Etot = 2.00) --> (Ekin = 1.50, Eint = 0.50):
//
//    (a) it is okay to apply the dual-energy fix when the output fluid has
//          (Dens = 0.10, MomX = 0.10, MomY = 0.10, MomZ = 0.10, Etot = 0.16) --> (Ekin = 0.15, Eint = 0.01),
//        where Eint/Ekin = 1./15. is assumed to be lower than DualEnergySwitch
//
//    (b) it is also okay to apply the dual-energy fix when the output fluid has a slightly negative Eint, such as
//          (Dens = 0.10, MomX = 0.10, MomY = 0.10, MomZ = 0.10, Etot = 0.12) --> (Ekin = 0.15, Eint = -0.03)
//
//    in these cases, the Etot will be fixed to physically correct Etot' = Ekin + (internal energy from DUAL)
//
//    however,
//    (c) if the output density is low, for example,
//          (Dens = 0.05, MomX = 0.10, MomY = 0.10, MomZ = 0.10, Etot = 0.15) --> (Ekin = 0.30, Eint = -0.15).
//        in this case, the dual-energy fixed Etot' = Ekin + (internal energy from DUAL) will become at least 2x
//
//    (d) it is worse if the output fluid has an even lower density while momentum remains finite,
//          (Dens = 1e-3, MomX = 0.10, MomY = 0.10, MomZ = 0.10, Etot = 0.15) --> (Ekin = 15.0, Eint = -14.85),
//        the dual-energy-fixed Etot' will be unphysically large due to the large kinetic energy
//        in this case, Etot' and Eint' are all positive and may pass the unphysical check
//
//    in these cases, we should leave the negative Eint to be handled by Flu_Close()
```

hyschive: - "in this case, the dual-energy fixed Etot'" --> "in this case, the dual-energy-fixed Etot'"
- "if the output density is low, for example," --> "if the output density is very low, the total energy updated by the dual-energy fix may increase significantly, for example,"
- Synchronize with the latest `main` branch.

hsinhaoHHuang: Updated.

PR #507, opened 2026-01-18
add a check for pressure in interpolation

Inhyeok-Song:
Adding a check for pressure in interpolation.

This can catch the failure when the pressure mode in EoS is stable, but not energy mode.

### DIFF
--- a/src/Interpolation/Interpolate.cpp
+++ b/src/Interpolation/Interpolate.cpp
@@ -401,9 +401,11 @@ void Interpolate_Iterate( real CData[], const int CSize[3], const int CStart[3],
                                                    ERROR_INFO, UNPHY_SILENCE )  )
                      Fail_ThisCell = true;
 
+#                 ifdef CHECK_E2P
                   if (  Hydro_IsUnphysical_Single( Pres, "interpolated pressure",        (real)0.0, HUGE_NUMBER,
                                                    ERROR_INFO, UNPHY_SILENCE )  )
                      Fail_ThisCell = true;
+#                 endif
                } // if ( EoS_DensPres2Eint_CPUPtr != NULL )
             } // if ( FData_is_Prim )
 


src/Interpolation/Interpolate.cpp
```diff
@@ -362,6 +362,7 @@ void Interpolate_Iterate( real CData[], const int CSize[3], const int CStart[3],
 
 //       5-3. additional check
          real Eint=NULL_REAL;
+         real Pres=NULL_REAL;
```
hyschive, line 370:
```suggestion
//        check the Eint --> Pres conversion for general EoS
#        if ( EOS != EOS_GAMMA  &&  EOS != EOS_COSMIC_RAY  &&  !defined BAROTROPIC_EOS )
#           define CHECK_E2P
#        endif
#        ifdef CHECK_E2P
         real Pres=NULL_REAL;
#        endif
```

src/Interpolation/Interpolate.cpp
```diff
@@ -383,12 +384,18 @@ void Interpolate_Iterate( real CData[], const int CSize[3], const int CStart[3],
 
                   Eint = EoS_DensPres2Eint_CPUPtr( Temp[DENS], Temp[ENGY], Passive,
                                                    EoS_AuxArray_Flt, EoS_AuxArray_Int, h_EoS_Table );
+                  Pres = EoS_DensEint2Pres_CPUPtr( Temp[DENS], Eint,       Passive,
+                                                   EoS_AuxArray_Flt, EoS_AuxArray_Int, h_EoS_Table );
```
hyschive, line 395: Enclose it within `#ifdef CHECK_E2P ... #endif`

src/Interpolation/Interpolate.cpp
```diff
@@ -383,12 +384,18 @@ void Interpolate_Iterate( real CData[], const int CSize[3], const int CStart[3],
...
 
 //                internal energy cannot be negative (even within machine precision) since a pressure floor has been applied
 //                when calling Hydro_Con2Pri()
                   if (  Hydro_IsUnphysical_Single( Eint, "interpolated internal energy", (real)0.0, HUGE_NUMBER,
                                                    ERROR_INFO, UNPHY_SILENCE )  )
                      Fail_ThisCell = true;
+
+                  if (  Hydro_IsUnphysical_Single( Pres, "interpolated pressure",        (real)0.0, HUGE_NUMBER,
+                                                   ERROR_INFO, UNPHY_SILENCE )  )
+                     Fail_ThisCell = true;
```
hyschive, line 407: Enclose it within `#ifdef CHECK_E2P ... #endif`

hyschive (commented):
@Inhyeok-Song Thanks for the contribution! It looks good and I only have one minor suggestion.

PR #521, opened 2026-04-27
Fix [Runtime-Parameters]—All wiki page rendering issue

### LINKED ISSUE #511: [Runtime-Parameters]—All wiki page takes too long to render
### 🎯 Goals

The wiki page [[Runtime-Parameters]—All](https://github.com/hyschive/gamer-fork/wiki/%5BRuntime-Parameters%5D-All) shows the following error message:
```
The wiki page took too long to render.
Please edit this wiki page’s content so it renders faster.
```
Solve this issue.

### 📋 Mandatory tasks

- [ ] See `Goals`

### Optional tasks

_No response_

### References

_No response_

### Supplementary info

_No response_

HsunYeong:
### Summary
Change link style in `tool/wiki/sync_runtime_parameter.py` to resolve #511.
### Changes
Change link style from `[[ parameter \| filename#parameter ]]` to `[parameter](filename#parameter)` for faster rendering.


hyschive:
@HsunYeong Thanks for this prompt fix! It works well on GitHub. I have a few questions/requests before merging.
- [x] Could you comment on why the revised syntax performs much faster?
- [x] The revised syntax doesn't seem to work with `gollum`. Is it possible to fix it? Please see [here](https://github.com/hyschive/gamer-fork/wiki/%5BContribute%5D-Update-Wiki#edit-wiki) about how to use `gollum` to display the wiki locally.
- [x] Should we apply the same fix across the entire wiki? I think this depends on whether `golllum` compatibility can be preserved. If so, we should update all relevant pages for consistency and improved performance. Otherwise, if the original _slower_ syntax is required for `gollum`, perhaps we should limit this change to `[Runtime-Parameters]-All.md`, since that appears to be the only large file that causes a rendering problem. What do you think?

@xuanweishan After Hsun-Yeong has addressed the above tasks, could you help review this PR? Please verify all updated links carefully. Thanks!

HsunYeong:
@hyschive 

- I think wiki link style (older syntax) is slower because it will check whether the target exists before generating the URL, while the markdown syntax will just provide the link regardlessly.
- It seems like `gollum` is trying to access to `[Runtime-Parameters]-XXX#XXX` instead of `%5BRuntime-Parameters%5D-XXX#XXX`, causing broken links. I will try to fix it.
- Wiki link style still has the advantage of simplicity while markdown links requires relative paths to be carefully set up and can easily cause broken links. For this case it is simple because the file is generated by script and everything is under the same directory. I'm not sure whether implement it to all other files will be as trivial.

hyschive:
> Wiki link style still has the advantage of simplicity while markdown links requires relative paths to be carefully set up and can easily cause broken links. For this case it is simple because the file is generated by script and everything is under the same directory. I'm not sure whether implement it to all other files will be as trivial.

OK. Then let's restrict this change to `[Runtime-Parameters]-All.md`

hyschive (approved):

xuanweishan:
I've checked the edited files and they seems good to me.

xuanweishan (approved):

PR #522, opened 2026-04-27
Bug fix for computing CPU socket number

koarakawaii:
* Use C++ Standard vector `SocketMask` and the mask method proposed by @hfhsieh (see discussion below) to compute the CPU socket number.

### DIFF
--- a/src/Auxiliary/Aux_GetCPUInfo.cpp
+++ b/src/Auxiliary/Aux_GetCPUInfo.cpp
@@ -1,4 +1,6 @@
 #include "GAMER.h"
+#include <algorithm>
+#include <vector>
 
 
 
@@ -24,7 +26,8 @@ void Aux_GetCPUInfo( const char *FileName )
    size_t len = 0;
    char String[2][MAX_STRING];
    char Trash[MAX_STRING];
-   int SocketNow = -1, SocketPrevious = -1;
+   int SocketNow;
+   std::vector<int> SocketList;
    int CorePerSocket = 0, NSocket = 0;
    bool GotFirstCPUInfo = false;
 
@@ -47,12 +50,15 @@ void Aux_GetCPUInfo( const char *FileName )
       if (  strcmp( String[0], "physical" ) == 0  &&  strcmp( String[1], "id" ) == 0 )
       {
          sscanf( line, "%s%s%s%d", String[0], String[1], Trash, &SocketNow );
-         if ( SocketNow != SocketPrevious )
-         {
-            SocketPrevious = SocketNow;
-            NSocket++;
-         }
+         SocketList.push_back(SocketNow);
       }
+      // Required for std::unique to work
+      std::sort(SocketList.begin(), SocketList.end());
+      // Moves unique elements to front, returns iterator to new end
+      auto last = std::unique(SocketList.begin(), SocketList.end());
+      // Remove trailing duplicates
+      SocketList.erase(last, SocketList.end());
+      NSocket = SocketList.size();
 
       if ( GotFirstCPUInfo )   continue;
 
--- a/src/Auxiliary/Aux_GetCPUInfo.cpp
+++ b/src/Auxiliary/Aux_GetCPUInfo.cpp
@@ -84,7 +84,7 @@ void Aux_GetCPUInfo( const char *FileName )
       }
    }
 
-   for (const auto& masked: SocketMask)
+   for ( const auto& masked: SocketMask )
    {
       if ( masked ) NSocket ++;
    }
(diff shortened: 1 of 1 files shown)


src/Auxiliary/Aux_GetCPUInfo.cpp
```diff
@@ -47,13 +50,18 @@ void Aux_GetCPUInfo( const char *FileName )
...
          sscanf( line, "%s%s%s%d", String[0], String[1], Trash, &SocketNow );
-         if ( SocketNow != SocketPrevious )
-         {
-            SocketPrevious = SocketNow;
-            NSocket++;
-         }
+         SocketList.push_back(SocketNow);
       }
 
+   // do sorting because std::unique removes consecutive duplicating elements
```
hfhsieh: I think sorting the array and erasing the duplicate in the `while` loop could be inefficient.

My idea is to use a boolean array (`SocketMask`) to track the ID, and count the unique socket after reading the `cpuinfo` file.
~~~ c
int SocketNow = -1; 
const int Max_Socket = 1024;
bool SocketMask[Max_Socket] = {0};

...

while ( getline(&line, &len, CPUInfo) != -1 )
{
   ...
      if (  strcmp( String[0], "physical" ) == 0  &&  strcmp( String[1], "id" ) == 0 )
      {
         sscanf( line, "%s%s%s%d", String[0], String[1], Trash, &SocketNow );
         SocketMask[SocketNow] = 1;
         }
      }
   ...
}

// count the unique sockets after the while loop
for (int i = 0; i < Max_Socket; i++) {
    if (SocketMask[i])   NSocket++;
}
~~~

koarakawaii: @hfhsieh : Sorry I did not notice that there is a `while` loop. I think we can either use your method, or move:
```
// do sorting because std::unique removes consecutive duplicating elements
   std::sort(SocketList.begin(), SocketList.end());
// moves unique elements to front, returns iterator to new end
   auto last = std::unique(SocketList.begin(), SocketList.end());
// remove trailing duplicates from last to end
   SocketList.erase(last, SocketList.end());
// obtain ultimate list size
   NSocket = SocketList.size();
```
out of the `while` loop. It seems to me applying which method (after the sorting is moved out of the `while` loop) is a problem of perferring saving more operations or saving more memory space (but neither of them seems to matter in this case, I guess). If you recommend using the bool mask approach, then I can remove this PR and we can apply your approach.

src/Auxiliary/Aux_GetCPUInfo.cpp
```diff
@@ -83,6 +82,15 @@ void Aux_GetCPUInfo( const char *FileName )
       }
    }
 
+// do sorting because std::unique removes consecutive duplicating elements
```
hfhsieh: Add spaces within the parentheses in the following lines, for example 

~~~ c
std::sort( SocketList.begin(), SocketList.end() );
~~~

src/Auxiliary/Aux_GetCPUInfo.cpp
hfhsieh: @koarakawaii 
Either approach is acceptable and meets our requirements. I will leave the final decision to @hyschive

koarakawaii: @hfhsieh : How about we use the mask method you suggested, but use a vector with flexible size (instead of fixing the array size to `1024`) such that we do not allocate redundant memory and not rely on the assumption of `Max_Socket`, while keeping the operation number low, as shown in the updated PR?

src/Auxiliary/Aux_GetCPUInfo.cpp
```diff
@@ -47,11 +49,10 @@ void Aux_GetCPUInfo( const char *FileName )
       if (  strcmp( String[0], "physical" ) == 0  &&  strcmp( String[1], "id" ) == 0 )
       {
          sscanf( line, "%s%s%s%d", String[0], String[1], Trash, &SocketNow );
-         if ( SocketNow != SocketPrevious )
-         {
-            SocketPrevious = SocketNow;
-            NSocket++;
-         }
+         if ( (SocketNow + 1) > SocketMask.size() )
+            SocketMask.resize(SocketNow+1, false);
```
hfhsieh: Add spaces within the parentheses

~~~ c
SocketMask.resize( SocketNow+1, false );
~~~

hfhsieh (commented):
Looks good to me. I left one minor comment.

src/Auxiliary/Aux_GetCPUInfo.cpp
koarakawaii: Revised~

hfhsieh (approved):

hyschive:
@koarakawaii Please click `Resolve conversation` for all remaining comments if this PR is ready for merging.

hyschive (approved):

PR #530, opened 2026-05-28
Add default GRACKLE_PATH and LIBYT_PATH to config files

HsunYeong:
Add default `GRACKLE_PATH` and `LIBYT_PATH` to `eureka_intel.config` and `spock_*.config`


configs/eureka_intel.config
```diff
@@ -4,9 +4,9 @@ FFTW2_PATH      /software/fftw/default
 FFTW3_PATH      /software/fftw/default3
 MPI_PATH        /software/openmpi/default
 HDF5_PATH       /software/hdf5/default
-GRACKLE_PATH
+GRACKLE_PATH    /software/grackle/3.4.2-dev-intel_double_omp
 GSL_PATH        /software/gsl/default
-LIBYT_PATH
+LIBYT_PATH      /software/libyt/0.5.0_omp
```
hyschive: Should we change it to `/software/libyt/0.5.0_omp_python3.12`?

hyschive (commented):
Thanks for the update! I have just one minor comment.

configs/eureka_intel.config
hyschive, line 1: Please also update `eureka_gnu.config`.

HsunYeong: I only add `LIBYT_PATH` since there is no gnu version of grackle.

hyschive (approved):

PR #533, opened 2026-06-26
Support finer data dump cadence for user-specified outputs

barrytchiang:
As part of #515, this PR implements support for finer-cadence data output that fires independently of
(and between) the main `Data_XXXXXX` dumps. We introduce `OPT__OUTPUT_SUBDIV` as a sub-cadence gate;
sub-cadence data are written to a single `SubData_%06d` file in the full-snapshot HDF5 format, whose
`Tree`, `GridData`, and `Particle` groups are controlled independently, from fully yt-loadable
snapshots down to minimal particles-only files for direct `h5py` access.

---

## Sub-cadence gate: `OPT__OUTPUT_SUBDIV`

**Convention:** `<0 = off` (default `-1`), `N >= 1 = enabled`. When enabled, at least one of
`OPT__OUTPUT_SUBDIV_GRID`/`_PAR`/`_TRACER`/`_USER` must be set.

- `N = 1`: fires at main dumps only; tracks `SubDumpID` and writes `Record__TimeSubDump`
- `N >= 2`: additionally fires `N-1` evenly spaced intermediate sub-dumps per main interval, plus at
  every main dump
  - `OPT__OUTPUT_MODE = 1`: sub-dump every `N` steps after each main dump
  - `OPT__OUTPUT_MODE = 2/3`: `N-1` evenly spaced sub-dumps per interval (dt is limited to hit
    `SubDumpTime` exactly)

---

## The `SubData_%06d` output file

A single file per sub-dump (indexed by `SubDumpID`) in the full-snapshot HDF5 format: `Info`,
`KeyInfo`, and `InputPara` are always written; the remaining groups are controlled independently:

| Group      | Flag | Default | Notes |
|------------|------|---------|-------|
| `Tree`     | `OPT__OUTPUT_SUBDIV_TREE` | 1 | required for yt loading; disable only to minimize file size for direct non-yt-based analysis |
| `GridData` | `OPT__OUTPUT_SUBDIV_GRID` | 0 | field list set by the (required) `Input__Sub_Grid`; supports native and derived fields for all models; an empty `GridData` group is still created when the tree is output without grid fields, since the yt gamer frontend requires the group |
| `Particle` | `OPT__OUTPUT_SUBDIV_PAR` / `OPT__OUTPUT_SUBDIV_TRACER` | 0/0 | massive and/or tracer particles, mixed in one patch-ordered array as in full snapshots (separate via `ParType` in post-processing); `OPT__OUTPUT_SUBDIV_FLOAT32` [0] downcasts the floating-point (incl. mesh-sampled) attributes to single precision |

`OPT__OUTPUT_SUBDIV_USER` [0] fires `Output_User_Ptr()` at sub-cadence (requires
`OPT__OUTPUT_USER = 1`); when enabled, `Output_User_Ptr()` fires exclusively via the sub-cadence
path (indexed by `SubDumpID`) at both sub-dumps and main dumps for consistent indexing.

**Field selection (`Input__Sub_Grid`):** one label per line; both native (conserved + passive)
fields and derived fields (`Pote`, `ParDens`/`TotalDens`, `CCMagX/Y/Z`, `Pres`, `Temp`, `Entr`,
`Cs`, `DivVel`, `Mach`, `DivMag`, the SRHD and GRACKLE fields, and user-defined derived fields) are
supported. Sub-dump derived fields are controlled solely by this list; the `OPT__OUTPUT_*` field
flags apply to main dumps only. An annotated template is provided at `example/input/Input__Sub_Grid`.

**Validation rules** (all fatal errors at initialization): `_GRID` requires `_TREE` and the file
`Input__Sub_Grid`; listing `ParDens`/`TotalDens` without `OPT__OUTPUT_PAR_DENS` (which selects the
deposition mode) or a user-defined derived field without `OPT__OUTPUT_USER_FIELD` is an error;
`OPT__OUTPUT_SUBDIV = 0` is forbidden.

**Type-filtered particle counts:** when only a subset of particle types is output,
`KeyInfo.Par_NPar` and the per-patch `Tree/NPar` are rewritten to the type-filtered counts so that
yt indexes the `Particle` datasets correctly (verified explicitly; see Validation).

---

## State tracking and restart safety

- A new global counter `SubDumpID` (distinct from `DumpID`) advances at every sub-dump event,
  including major dumps when any sub-cadence flag is enabled.
- A new record file `Record__TimeSubDump` (columns: `SubDumpID`, `Time`, `Step`) is written
  alongside the existing `Record__Dump`.
- When `OPT__OUTPUT_MODE = 2/3`, dt limiting in `Mis_GetTimeStep` ensures the simulation hits
  `SubDumpTime` exactly.
- `SubDumpID` is stored in the HDF5 `KeyInfo` group and restored (`+1`) on restart; `LoadField` is
  called with `NonFatal` for backward compatibility with pre-feature HDF5 files. `SubDumpTime` needs
  no storage since it is always re-derived at initialization.
- All seven `OPT__OUTPUT_SUBDIV*` parameters are recorded in `Record__Note` and the `InputPara`
  HDF5 group; `Record__TimeSubDump` and `Sub*` patterns are added to all `clean.sh` files.

---

## Code organization

Sub-dump output is fully unified into the full-snapshot writer: `Output_DumpData_Total_HDF5()`
takes a single `SubDataMode` flag and both sub-dump call sites collapse to one writer invocation
(the former separate compact particle writer `Par_Output_SubDump.cpp` is removed). New source
files: `src/Init/Init_SubGrid_Fields.cpp` (parses and validates `Input__Sub_Grid`; provides
`SubGrid_DerFieldSelected()`) and `example/input/Input__Sub_Grid`.

---

## Validation

1. Full configuration (Plummer example, which presets `_GRID/_PAR/_TREE = 1` with its own
   `Input__Sub_Grid`, exercised by simply setting `OPT__OUTPUT_SUBDIV = 2`): `SubData_*` contains
   exactly the selected fields (including derived `Pote` chosen purely via the list), the complete
   `Particle` group, and consistent counts (`KeyInfo.Par_NPar` = `Tree/NPar` sum = yt particle
   count); intermediate sub-dumps land exactly on the half-interval; every file loads through the
   standard yt gamer frontend.
2. Minimal configuration (`_TREE/_GRID = 0`, `_FLOAT32 = 1`): the file reduces to
   `Info`/`Particle`/`User` with float32 particle data, essentially matching the size of a bare
   particle dump.
3. Type-filtered configuration (ParticleTest with mixed massive+tracer particles, `_TRACER = 1`,
   `_PAR = 0`): counts correctly reflect the tracers-only selection and yt loads the file natively.
4. The new `example/test_problem/Hydro/Plummer/plot_script/plot_subdata.py` produces the
   grid+particle visualization from `SubData_*` alone.
5. Zeldovich (`_USER` mode): the sub-cadence `Output_User_Ptr()` text outputs and the analysis
   script's automatic `DumpID` to `SubDumpID` remapping work unchanged, and a pre-existing
   `Input__Parameter` without the new parameters runs correctly on their defaults (backward
   compatibility of old input files).
6. Compile matrix: HYDRO (production flags), ELBDM, MHD, SRHD, and GRACKLE configurations.

## Documentation and example updates

- Wiki: the seven sub-cadence parameters documented in `[Runtime-Parameters]-Outputs` (and the
  auto-synced `[Runtime-Parameters]-All` table via `tool/wiki/sync_runtime_parameter.py`), a
  `[Simulation-Logs]-Record__TimeSubDump` page, and the sub-cadence output description in `Outputs`.
- All 59 example `Input__Parameter` files updated with the new seven-parameter block.
- `example/test_problem/Hydro/Plummer` is the hydro+particle reference example (presets +
  `Input__Sub_Grid` + `plot_subdata.py`); `example/test_problem/Hydro/Zeldovich` remains the
  `_USER`-mode reference with automatic `DumpID` to `SubDumpID` remapping in its analysis script.


example/test_problem/Hydro/Zeldovich/plots/plot_analysis_Zeldovich.py
```diff
@@ -118,6 +136,7 @@ def Subplots_Vertical_Combine(input_filename_1,input_filename_2,output_filename)
 for ds in ts.piter():
     # load data frame cosmological parameters
     DumpID_now       = ds.parameters["DumpID"]
+    FileID_now       = get_file_id( DumpID_now )               # SubDumpID for Zeldovich_* files (= DumpID when OPT__OUTPUT_SUBDIV < 0)
```
HsunYeong:
```suggestion
    FileID_now       = get_file_id( DumpID_now )                # SubDumpID for Zeldovich_* files (= DumpID when OPT__OUTPUT_SUBDIV < 0)
```

src/Init/Init_ByRestart_HDF5.cpp
```diff
@@ -201,6 +201,8 @@ void Init_ByRestart_HDF5( const char *FileName )
    MPI_Barrier( MPI_COMM_WORLD );
 
    LoadField( "DumpID",               &KeyInfo.DumpID,               H5_SetID_KeyInfo, H5_TypeID_KeyInfo,    Fatal,  NullPtr,              -1, NonFatal );
+   LoadField( "SubDumpID",           &KeyInfo.SubDumpID,           H5_SetID_KeyInfo, H5_TypeID_KeyInfo, NonFatal,  NullPtr,              -1, NonFatal );
+   LoadField( "SubDumpTime",          &KeyInfo.SubDumpTime,          H5_SetID_KeyInfo, H5_TypeID_KeyInfo, NonFatal,  NullPtr,              -1, NonFatal );
```
HsunYeong:
```suggestion
   LoadField( "SubDumpID",            &KeyInfo.SubDumpID,            H5_SetID_KeyInfo, H5_TypeID_KeyInfo, NonFatal,  NullPtr,              -1, NonFatal );
   LoadField( "SubDumpTime",          &KeyInfo.SubDumpTime,          H5_SetID_KeyInfo, H5_TypeID_KeyInfo, NonFatal,  NullPtr,              -1, NonFatal );
```

src/Auxiliary/Aux_Check_Parameter.cpp
```diff
@@ -387,12 +387,59 @@ void Aux_Check_Parameter()
...
+                 OPT__OUTPUT_SUBDIV_PAR );
+
+   if ( OPT__OUTPUT_SUBDIV_TRACER < 0  ||  OPT__OUTPUT_SUBDIV_TRACER > 2 )
+      Aux_Error( ERROR_INFO, "OPT__OUTPUT_SUBDIV_TRACER = %d is not supported; valid values: 0=off, 1=on, 2=on+float32 !!\n",
+                 OPT__OUTPUT_SUBDIV_TRACER );
+
+// any sub-cadence output flag requires OPT__OUTPUT_SUBDIV >= 1;
+// conversely, OPT__OUTPUT_SUBDIV >= 1 requires at least one flag to be enabled
+   const bool AnySubDiv = OPT__OUTPUT_SUBDIV_GRID > 0  ||  OPT__OUTPUT_SUBDIV_PAR  ||
+                          OPT__OUTPUT_SUBDIV_TRACER  ||  OPT__OUTPUT_SUBDIV_USER;
```
HsunYeong:
```suggestion
   const bool AnySubDiv = OPT__OUTPUT_SUBDIV_GRID    ||  OPT__OUTPUT_SUBDIV_PAR  ||
                          OPT__OUTPUT_SUBDIV_TRACER  ||  OPT__OUTPUT_SUBDIV_USER;
```
More consistent format.

src/Auxiliary/Aux_Check_Parameter.cpp
```diff
@@ -387,12 +387,59 @@ void Aux_Check_Parameter()
...
+
+#  ifndef PARTICLE
+   if ( OPT__OUTPUT_SUBDIV_PAR )
+      Aux_Error( ERROR_INFO, "OPT__OUTPUT_SUBDIV_PAR requires PARTICLE to be enabled at compile time !!\n" );
+   if ( OPT__OUTPUT_SUBDIV_TRACER )
+      Aux_Error( ERROR_INFO, "OPT__OUTPUT_SUBDIV_TRACER requires PARTICLE (TRACER) to be enabled at compile time !!\n" );
+#  endif
+
+#  ifndef SUPPORT_HDF5
+   if ( OPT__OUTPUT_SUBDIV_GRID > 0  ||  OPT__OUTPUT_SUBDIV_PAR  ||  OPT__OUTPUT_SUBDIV_TRACER )
```
HsunYeong:
```suggestion
   if ( OPT__OUTPUT_SUBDIV_GRID  ||  OPT__OUTPUT_SUBDIV_PAR  ||  OPT__OUTPUT_SUBDIV_TRACER )
```
More consistent format

src/Miscellaneous/Mis_GetTimeStep.cpp
```diff
@@ -155,6 +155,27 @@ double Mis_GetTimeStep( const int lv, const double dTime_SyncFaLv, const double
       sprintf( dTime_Name[NdTime++], "%s", "Data_Dump" );
    }
 
+// 1.4.5 CRITERION : match the time of the next sub-dump (only needed for intermediate sub-dumps, i.e., OPT__OUTPUT_SUBDIV>=2)
+   const bool SubDumpByTime = ( OPT__OUTPUT_SUBDIV >= 2  &&  OPT__OUTPUT_USER  &&  DumpByTime );
```
HsunYeong:
```suggestion
   const bool SubDumpByTime = ( OPT__OUTPUT_SUBDIV >= 2  &&  DumpByTime );
```
`OPT__OUTPUT_USER` should not be a requirement anymore since the implementation of more general output fields.

src/Output/Output_DumpData_Total_HDF5.cpp
```diff
@@ -288,7 +288,7 @@ Procedure for outputting new variables:
 //                                             DT__GRACKLE_COOLING, OPT__FLAG_COOLING_LEN, FlagTable_CoolingLen
 //                2508 : 2026/03/26 --> output particle unique id
 //-------------------------------------------------------------------------------------------------------
-void Output_DumpData_Total_HDF5( const char *FileName )
+void Output_DumpData_Total_HDF5( const char *FileName, const bool SkipPar, const int SubGridMode )
 {
 
    if ( MPI_Rank == 0 )    Aux_Message( stdout, "%s (DumpID = %d)     ...\n", __FUNCTION__, DumpID );
```
HsunYeong: Add a way to distinguish regular dump and subdump here? For subdump it should print:
` Aux_Message( stdout, "%s (SubDumpID = %d)     ...\n", __FUNCTION__, SubDumpID );`

src/Output/Output_DumpData.cpp
```diff
@@ -296,17 +304,118 @@ void Output_DumpData( const int Stage )
 
       Write_DumpRecord();
 
+//    fire sub-cadence outputs at N=1 (or at every main dump when SUBDIV >= 1)
+      const bool AnySubDiv = OPT__OUTPUT_SUBDIV_GRID > 0  ||  OPT__OUTPUT_SUBDIV_PAR  ||
+                             OPT__OUTPUT_SUBDIV_TRACER  ||  OPT__OUTPUT_SUBDIV_USER;
```
HsunYeong:
```suggestion
      const bool AnySubDiv = OPT__OUTPUT_SUBDIV_GRID    ||  OPT__OUTPUT_SUBDIV_PAR  ||
                             OPT__OUTPUT_SUBDIV_TRACER  ||  OPT__OUTPUT_SUBDIV_USER;
```
More consistent format.

src/Output/Output_DumpData.cpp
```diff
@@ -296,17 +304,118 @@ void Output_DumpData( const int Stage )
...
+      if ( AnySubDiv  &&  OPT__OUTPUT_SUBDIV >= 1 )
+      {
+         char SubFileName[2*MAX_STRING];
+#        ifdef SUPPORT_HDF5
+         if ( OPT__OUTPUT_SUBDIV_GRID > 0 )
```
HsunYeong:
```suggestion
         if ( OPT__OUTPUT_SUBDIV_GRID )
```
More consistent format.

src/Output/Output_DumpData.cpp
```diff
@@ -296,17 +304,118 @@ void Output_DumpData( const int Stage )
...
       }
 
       PreviousDumpStep = Step;
    } // if ( OutputData || OutputData_RunTime )
 
+
+// sub-dump gate: fire individual sub-cadence outputs between main dumps
+   {
+      const bool AnySubDiv2 = OPT__OUTPUT_SUBDIV_GRID > 0  ||  OPT__OUTPUT_SUBDIV_PAR  ||
+                              OPT__OUTPUT_SUBDIV_TRACER  ||  OPT__OUTPUT_SUBDIV_USER;
```
HsunYeong:
```suggestion
      const bool AnySubDiv2 = OPT__OUTPUT_SUBDIV_GRID    ||  OPT__OUTPUT_SUBDIV_PAR  ||
                              OPT__OUTPUT_SUBDIV_TRACER  ||  OPT__OUTPUT_SUBDIV_USER;
```
More consistent format.

src/Output/Output_DumpData.cpp
```diff
@@ -296,17 +304,118 @@ void Output_DumpData( const int Stage )
...
+
+            default :
+               Aux_Error( ERROR_INFO, "incorrect parameter %s = %d !!\n", "OPT__OUTPUT_MODE", OPT__OUTPUT_MODE );
+         } // switch ( OPT__OUTPUT_MODE )
+
+         if ( SubDump )
+         {
+            char SubFileName[2*MAX_STRING];
+#           ifdef SUPPORT_HDF5
+            if ( OPT__OUTPUT_SUBDIV_GRID > 0 )
```
HsunYeong:
```suggestion
            if ( OPT__OUTPUT_SUBDIV_GRID )
```
More consistent format.

src/Init/Init_ByRestart_HDF5.cpp
```diff
@@ -353,6 +355,13 @@ void Init_ByRestart_HDF5( const char *FileName )
    else
       DumpID = INIT_DUMPID;
 
+// 1-8b. recover user sub-dump state
+   if ( ! OPT__RESTART_RESET )
+   {
+      SubDumpID  = KeyInfo.SubDumpID;
```
HsunYeong:
```suggestion
      SubDumpID   = KeyInfo.SubDumpID + 1;
```
Without `+1` the last subdump before restart will be overwritten.

src/Output/Output_DumpData.cpp
```diff
@@ -95,6 +100,9 @@ void Output_DumpData( const int Stage )
          break;
 
       } // switch ( OPT__OUTPUT_MODE )
+
+//    for restarts: SubDumpTime and SubDumpID are recovered from HDF5 in Init_ByRestart_HDF5;
+//    for fresh starts: SubDumpTime is set after the Stage==0 main dump fires (see below)
```
HsunYeong:
```suggestion

//    initialize sub-dump cadence
      if ( OPT__OUTPUT_SUBDIV >= 2 )
      {
         LastMainDumpStep = Step;

         if ( OPT__OUTPUT_MODE == OUTPUT_CONST_DT || OPT__OUTPUT_MODE == OUTPUT_USE_TABLE )
         {
            SubInterval = ( DumpTime - Time[0] ) / OPT__OUTPUT_SUBDIV;
            SubDumpTime = Time[0] + SubInterval;
         }
      }

```
Initialize sub-dump cadence at stage 0.
For `OPT__INIT == INIT_BY_RESTART  &&  Stage == 0`, `SubDumpTime` and `SubInterval` won't be properly initialized due to quick return of this function, and carrying `SubDumpTime` from restart file will cause too small time step when computing `dTime[NdTime] = SubDumpTime - Time[lv]` in `Mis_GetTimeStep()`(since `SubDumpTime = Time[0]`).

src/Particle/Par_Output_SubDump.cpp
```diff
@@ -0,0 +1,263 @@
...
+static void Par_Output_SubTracer_Internal( const char *FileName, const int SubDumpID )
+{
+   Par_Output_SubDump_HDF5( FileName, true, SubDumpID );
+}
+#endif
+
+static void Par_Output_SubParticle_Internal( const char *FileName, const int SubDumpID )
+{
+   Par_Output_SubDump_HDF5( FileName, false, SubDumpID );
+}
```
HsunYeong: Maybe put these inside `Par_Output_SubTracer()` and `Par_Output_SubParticle()`?

src/Output/Output_DumpData_Total_HDF5.cpp
```diff
@@ -324,6 +328,23 @@ void Output_DumpData_Total_HDF5( const char *FileName )
       }
 #     endif
 
+//    field masking for grid sub-dumps (SubGridMode 2-4); SubGridMode 0/1 writes all fields
```
HsunYeong: The masking only works for primitive variables. For example, if `SubGridMode == 2` and `OPT__OUTPUT_PRES == 1` the sub grid data will have both `Dens` and `Pres`. Maybe we can let user choose different field combinations by providing a input file similar to `Input_Par_Mesh`, what do you think?

HsunYeong (commented):
@barrytchiang Thank you for your contribution! I just added some comments.

src/Init/Init_Load_Parameter.cpp
```diff
@@ -572,7 +572,7 @@ void Init_Load_Parameter()
    ReadPara->Add( "OUTPUT_DT",                  &OUTPUT_DT,                      -1.0,             NoMin_double,  NoMax_double   );
 // do not check OPT__OUTPUT_SUBDIV range here since validation is in Aux_Check_Parameter (0 is forbidden; <0=off, >=1=on)
    ReadPara->Add( "OPT__OUTPUT_SUBDIV",         &OPT__OUTPUT_SUBDIV,             -1,               NoMin_int,     NoMax_int      );
-   ReadPara->Add( "OPT__OUTPUT_SUBDIV_GRID",    &OPT__OUTPUT_SUBDIV_GRID,        0,                0,             NoMax_int      );
+   ReadPara->Add( "OPT__OUTPUT_SUBDIV_GRID",    &OPT__OUTPUT_SUBDIV_GRID,    false,            false,             true           );
```
HsunYeong:
```suggestion
   ReadPara->Add( "OPT__OUTPUT_SUBDIV_GRID",    &OPT__OUTPUT_SUBDIV_GRID,        false,            Useless_bool,  Useless_bool   );
```

src/Main/Main.cpp
```diff
@@ -54,7 +54,9 @@ double               BOX_SIZE, DT__MAX, DT__FLUID, DT__FLUID_INIT, END_T, OUTPUT
 long                 END_STEP;
 int                  NX0_TOT[3], OUTPUT_STEP, OUTPUT_WALLTIME_UNIT, REGRID_COUNT, REFINE_NLEVEL, FLU_GPU_NPGROUP, SRC_GPU_NPGROUP, OMP_NTHREAD;
 int                  OPT__OUTPUT_SUBDIV             = -1;
-int                  OPT__OUTPUT_SUBDIV_GRID        = 0;
+bool                 OPT__OUTPUT_SUBDIV_GRID        = false;
+bool                 SubGridField[NCOMP_TOTAL]       = {};   // zero-initialized = all fields excluded until Init_SubGrid_Fields()
+int                  SubGridField_Num                = 0;
```
HsunYeong:
```suggestion
bool                 OPT__OUTPUT_SUBDIV_GRID        = false;
bool                 SubGridField[NCOMP_TOTAL]      = {};   // zero-initialized = all fields excluded until Init_SubGrid_Fields()
int                  SubGridField_Num               = 0;
```

example/test_problem/Template/Input__Parameter
```diff
@@ -420,10 +420,10 @@ OPT__OUTPUT_GRACKLE_TCOOL     0           # output the cooling time calculated b
 OPT__OUTPUT_USER_FIELD        0           # output user-defined derived fields [0] -> edit "Flu_DerivedField_User.cpp"
 OPT__OUTPUT_MODE              1           # (1=const step, 2=const dt, 3=dump table) -> edit "Input__DumpTable" for 3
 OPT__OUTPUT_SUBDIV           -1           # sub-cadence gate: <0=off; N>=1: sub-dumps between/at main dumps [-1]
-                                          # for OPT__OUTPUT_MODE == 1, sub-dump every N steps after each main dump (e.g., N=3, const_step=7: sub-dumps at steps 0, 3, 6, 7, 10, 13, 14,...)
+                                          # for OPT__OUTPUT_MODE == 1, sub-dump every N steps after each main dump
```
HsunYeong:
```suggestion
                                          # for OPT__OUTPUT_MODE == 1, sub-dump every N steps after each main dump
```
There is a trailing space for all `Input__Parameter` files.

src/Auxiliary/Aux_Check_Parameter.cpp
```diff
@@ -388,7 +388,7 @@ void Aux_Check_Parameter()
       Aux_Message( stderr, "WARNING : OPT__OUTPUT_TOTAL = 2 (C-binary) is deprecated !!\n" );
 
    if ( !OPT__OUTPUT_TOTAL  &&  !OPT__OUTPUT_PART  &&  !OPT__OUTPUT_USER  &&  !OPT__OUTPUT_BASEPS  &&
```
HsunYeong: Align `&&`s.

src/Auxiliary/Aux_Check_Parameter.cpp
```diff
@@ -411,7 +407,7 @@ void Aux_Check_Parameter()
 
 // any sub-cadence output flag requires OPT__OUTPUT_SUBDIV >= 1;
 // conversely, OPT__OUTPUT_SUBDIV >= 1 requires at least one flag to be enabled
-   const bool AnySubDiv = OPT__OUTPUT_SUBDIV_GRID > 0  ||  OPT__OUTPUT_SUBDIV_PAR  ||
+   const bool AnySubDiv = OPT__OUTPUT_SUBDIV_GRID  ||  OPT__OUTPUT_SUBDIV_PAR  ||
                           OPT__OUTPUT_SUBDIV_TRACER  ||  OPT__OUTPUT_SUBDIV_USER;
```
HsunYeong:
```suggestion
   const bool AnySubDiv = OPT__OUTPUT_SUBDIV_GRID    ||  OPT__OUTPUT_SUBDIV_PAR  ||
                          OPT__OUTPUT_SUBDIV_TRACER  ||  OPT__OUTPUT_SUBDIV_USER;
```

src/Output/Output_DumpData.cpp
```diff
@@ -284,37 +295,40 @@ void Output_DumpData( const int Stage )
...
+      }
 
-//    fire sub-cadence outputs at N=1 (or at every main dump when SUBDIV >= 1)
-      const bool AnySubDiv = OPT__OUTPUT_SUBDIV_GRID > 0  ||  OPT__OUTPUT_SUBDIV_PAR  ||
+//    fire sub-cadence outputs at N=1 (or at every main dump when SUBDIV >= 1) BEFORE writing
+//    the main HDF5 snapshot so that SubDumpID is already incremented when the snapshot is written;
+//    without this, restarting from a main-dump snapshot restores the pre-increment SubDumpID and
+//    the first sub-dump after restart collides (same filename) with the co-dump at the main dump
+      const bool AnySubDiv = OPT__OUTPUT_SUBDIV_GRID  ||  OPT__OUTPUT_SUBDIV_PAR  ||
                              OPT__OUTPUT_SUBDIV_TRACER  ||  OPT__OUTPUT_SUBDIV_USER;
```
HsunYeong:
```suggestion
      const bool AnySubDiv = OPT__OUTPUT_SUBDIV_GRID    ||  OPT__OUTPUT_SUBDIV_PAR  ||
                             OPT__OUTPUT_SUBDIV_TRACER  ||  OPT__OUTPUT_SUBDIV_USER;
```

src/Output/Output_DumpData.cpp
```diff
@@ -333,33 +347,35 @@ void Output_DumpData( const int Stage )
...
 
       PreviousDumpStep = Step;
    } // if ( OutputData || OutputData_RunTime )
 
 
 // sub-dump gate: fire individual sub-cadence outputs between main dumps
    {
-      const bool AnySubDiv2 = OPT__OUTPUT_SUBDIV_GRID > 0  ||  OPT__OUTPUT_SUBDIV_PAR  ||
+      const bool AnySubDiv2 = OPT__OUTPUT_SUBDIV_GRID  ||  OPT__OUTPUT_SUBDIV_PAR  ||
                               OPT__OUTPUT_SUBDIV_TRACER  ||  OPT__OUTPUT_SUBDIV_USER;
```
HsunYeong:
```suggestion
      const bool AnySubDiv2 = OPT__OUTPUT_SUBDIV_GRID    ||  OPT__OUTPUT_SUBDIV_PAR  ||
                              OPT__OUTPUT_SUBDIV_TRACER  ||  OPT__OUTPUT_SUBDIV_USER;
```

src/Output/Output_DumpData.cpp
```diff
@@ -284,37 +295,40 @@ void Output_DumpData( const int Stage )
...
       if ( Output_UserWorkBeforeOutput_Ptr != NULL )  Output_UserWorkBeforeOutput_Ptr();
 
-//    start dumping data
-      if ( OPT__OUTPUT_TOTAL )            Output_DumpData_Total( FileName_Total );
-      if ( OPT__OUTPUT_PART  )            Output_DumpData_Part( OPT__OUTPUT_PART, OPT__OUTPUT_BASE, OUTPUT_PART_X,
-                                                                OUTPUT_PART_Y, OUTPUT_PART_Z, FileName_Part );
-      if ( OPT__OUTPUT_USER  &&  !OPT__OUTPUT_SUBDIV_USER )
+//    advance DumpTime and SubDumpTime before writing the HDF5 so that the stored
+//    SubDumpTime is already the next sub-dump target; without this, restarting from
+//    a main-dump snapshot gives SubDumpTime == Time[0] → dTime == 0 → crash
```
HsunYeong: I think we should not record the next `SubDumpTime` which may lead to confusion. Since `SubDumpTime` will always be initialized during `Stage==0` now, I think there is actually no need to record `SubDumpTime` in HDF5 data. What do you think?

src/Output/Output_DumpData_Total_HDF5.cpp
```diff
@@ -324,6 +332,15 @@ void Output_DumpData_Total_HDF5( const char *FileName )
       }
 #     endif
 
+//    field masking for grid sub-dumps; field list controlled by SubGridField[] (Input__Sub_Grid)
+#     if ( MODEL == HYDRO )
+      if ( SubGridMode  &&  !SubGridField[v] ) { NCompFluSkip += 1; continue; }
+#     endif
```
HsunYeong: I think there is no need to set this as `HYDRO` only

src/Init/Init_SubGrid_Fields.cpp
HsunYeong, line 1: I'm wondering if we should also support derived fields to give user more controls. For example, 12 derived fields defined in `Macro.h` for `MODEL == HYDRO`, which is also supported by `Input__Par_Mesh`.

example/input/Input__Sub_Grid
```diff
@@ -0,0 +1,16 @@
+# List of native fluid field labels for grid sub-dumps (OPT__OUTPUT_SUBDIV_GRID = 1).
+# One label per non-comment line.  The input label is case-sensitive.
+#
+# Supported labels (HYDRO):
+#   Conserved  : Dens, MomX, MomY, MomZ, Engy
+#   Passive    : any user-defined passive scalar (check Init_Field.cpp for labels)
+#
```
HsunYeong: I think it can support ELBDM fields if we remove the `#if ( MODEL == HYDRO )` condition in `Output/Output_DumpData_Total_HDF5.cpp` (see my comment on that file).

HsunYeong (commented):
@barrytchiang Thank you for your update! I just added a few minor comments.

example/test_problem/Template/Input__Parameter
barrytchiang: Good catch — will sweep the trailing space across all example Input__Parameter files in the next commit.

src/Auxiliary/Aux_Check_Parameter.cpp
barrytchiang: Will fix in the next commit.

example/input/Input__Sub_Grid
barrytchiang: Right — will update the header notes together with removing the HYDRO guard.

src/Init/Init_SubGrid_Fields.cpp
barrytchiang: Good idea — will add derived-field support to Input__Sub_Grid mirroring the Input__Par_Mesh treatment in the next commit.

src/Output/Output_DumpData_Total_HDF5.cpp
barrytchiang: Agreed — will remove the MODEL == HYDRO restriction.

src/Output/Output_DumpData.cpp
barrytchiang: Agreed — with the Stage-0 initialization, SubDumpTime is always recomputed on restart and the stored value is never consumed. Will remove SubDumpTime from KeyInfo (and the pre-advance logic here) in the next commit; SubDumpID stays.

HsunYeong (approved):
@barrytchiang Thanks for the update! The PR looks good to me now.

HsunYeong:
@barrytchiang Some additional comments:

- Merge `Par_Output_SubDump()` into `Output_DumpData_Total_HDF5()` to reduce code duplication.
- Maybe separate `OPT__OUTPUT_POT`, etc and `Input__SubGrid` table? Make `OPT__OUTPUT_*` specifically for main dump, with something like 
`OutPot = ( SubGridMode )? SubGrid_DerFieldSelected(PotLabel) : OPT__OUTPUT_POT;`
- Update wiki.
- Setup default `OPT__OUTPUT_SUBDIV_*` for a test problem with both Hydro and particle (e.g. Plummer), so one can easily test this function by simply enabling `OPT__OUTPUT_SUBDIV`.
- file a issue/to-do list to specify `OPT__OUTPUT_PART`, `OPT__OUTPUT_PAR_MODE`, `OPT__OUTPUT_BASEPS` are not supported by sub-dump cadence yet (low priority).

barrytchiang:
@HsunYeong 
Thanks for the additional comments! Before implementing the `Par_Output_SubDump()` merge, I'd like to clarify the intended scope, since the two particle write paths are structurally quite different:

- The main-dump path in `Output_DumpData_Total_HDF5()` writes particles **patch-ordered per level**  (via the LB/GID machinery) into the `Particle` group of `Data_XXXXXX`.
- The sub-dump path writes a **flat, type-filtered** (tracer vs. massive) list via rank-by-rank hyperslab writes into a compact standalone file, with optional float32 downcast and mesh-sampled attributes for tracers.

The genuinely shared code is mostly the HDF5 plumbing (dataset creation, hyperslab selection/write loop) rather than the data marshaling itself. So I see two options:

**(a)** Relocate `Par_Output_SubParticle()`/`Par_Output_SubTracer()` into
`Output_DumpData_Total_HDF5.cpp` (removing `Par_Output_SubDump.cpp`) and factor the common HDF5 plumbing into shared static helpers used by both paths. Low risk; the main-dump path is untouched and both file formats stay bit-identical.
**(b)** Fully unify the two particle write paths behind one generalized routine. This is a deeper refactor that would touch the main-dump particle output, which I'd prefer not to destabilize in this PR — if you'd like (b), I'd suggest folding it into the low-priority to-do issue (together with the `OPT__OUTPUT_PART`/`OPT__OUTPUT_PAR_MODE`/`OPT__OUTPUT_BASEPS` item) and doing it as a follow-up PR.

Which did you have in mind? I'll proceed with the other items (`Input__Sub_Grid` decoupling with `OPT__OUTPUT_*` as main-dump-only, wiki update, Plummer sub-dump defaults, and filing the to-do issue) in the meantime.

One note on the `OutPot = ( SubGridMode ) ? SubGrid_DerFieldSelected(PotLabel) : OPT__OUTPUT_POT` suggestion: I'll keep the absent-`Input__Sub_Grid` fallback mapping to the `OPT__OUTPUT_*` flags (i.e., `( SubGridMode && !SubGridDerAll ) ? ... : ...`), since with the file absent `SubGrid_DerFieldSelected()` returns true for every label and sub-dumps would otherwise write *all* derived fields regardless of the flags. With the file present, sub-dumps are then controlled solely by `Input__Sub_Grid`, exactly as you proposed. Two further small exceptions I'll document: `ParDens`/`TotalDens` still requires `OPT__OUTPUT_PAR_DENS` (the flag selects the deposition mode), and user-defined derived fields still require `OPT__OUTPUT_USER_FIELD` (their labels are only initialized when it is enabled).

*While updating the wiki I ran `tool/wiki/sync_runtime_parameter.py` locally to regenerate the `[Runtime-Parameters]-All` page, and it reports three pre-existing parameters with no description on any runtime-parameter wiki page: `SPEC_INT_XY_INSTEAD_DEPHA`, `SPEC_INT_VORTEX_THRESHOLD`, and `SPEC_INT_GHOST_BOUNDARY` (all ELBDM spectral-interpolation options, unrelated to this PR). Just flagging it.

HsunYeong:
@barrytchiang Thank for your detailed explanation!
- For particle dump I prefer (a) method, keeping the format can avoid the potential incompatibility with existing analysis scripts. I think we only need to apply filter if there is only one of `OPT__OUTPUT_SUBDIV_PAR` and `OPT__OUTPUT_SUBDIV_TRACER` is turned on.
- How about returning an error message when `OPT__OUTPUT_SUBDIV_GRID=1` and there is no `Input__Sub_Grid`?
- I agree with you about the `OPT__OUTPUT_PAR_DENS` and `OPT__OUTPUT_USER_FIELD` part.

hyschive:
> - The main-dump path in Output_DumpData_Total_HDF5() writes particles patch-ordered per level
(via the LB/GID machinery) into the Particle group of Data_XXXXXX.
> - The sub-dump path writes a flat, type-filtered (tracer vs. massive) list via rank-by-rank
hyperslab writes into a compact standalone file, with optional float32 downcast and mesh-sampled
attributes for tracers.

@barrytchiang @HsunYeong I thought the gamer frontend in yt assumes that particle data are stored in patch order. Are you sure the same yt script can analyze both the "main-dump" and "sub-dump" data formats?

barrytchiang:
@HsunYeong Sounds good! I'll proceed with (a), keeping both sub-dump file formats bit-identical!

One quick confirmation on the filter refinement: when both `OPT__OUTPUT_SUBDIV_PAR` and `OPT__OUTPUT_SUBDIV_TRACER` are on, do you mean writing a single combined particle file with no type filtering (as in the main dumps), and only applying the tracer/massive filter when exactly one of the two flags is enabled? If so, two small details I'd like to confirm: (i) the combined file would be named `SubParticle_%06d` (with `SubTracer_%06d` produced only in the tracer-only case), and (ii) since the two flags carry independent float32 settings (`= 2`), I'd require them to match when both are enabled (validated in `Aux_Check_Parameter()`), or should the combined file simply follow `OPT__OUTPUT_SUBDIV_PAR`?

Returning an error when `OPT__OUTPUT_SUBDIV_GRID = 1` and `Input__Sub_Grid` is absent: agreed, and it's a nice simplification: the absent-file fallback (`SubGridDerAll`) disappears entirely, and the derived-field gates reduce to exactly your original suggestion, `OutPot = ( SubGridMode ) ? SubGrid_DerFieldSelected(PotLabel) : OPT__OUTPUT_POT`. I'll also add an `Input__Sub_Grid` to the Plummer example accordingly.


@hyschive You're right: the gamer frontend in yt indexes the `Particle` group through the per-patch particle counts in `Tree`, so it cannot read the sub-dump particle files. That is by design: `SubParticle_*`/`SubTracer_*` are compact, flat, self-describing HDF5 files intended for direct `h5py`/NumPy access (each attribute is a named 1D dataset), which is what makes the high-cadence tracer analysis cheap; they are not yt-loadable and are not meant to be read by the same scripts as `Data_XXXXXX`. The yt-compatible sub-dump output is the grid-only `SubGrid_*` file: it is written by the same `Output_DumpData_Total_HDF5()` machinery as the main snapshots (with particles skipped), and we have verified explicitly that yt loads it through the standard gamer frontend (`GAMERDataset`, full hierarchy and field access, including derived fields selected via `Input__Sub_Grid`). I'll tighten the PR description to state this split explicitly. Making the particle sub-dumps yt-loadable would require a patch-ordered layout plus the `Tree` metadata (essentially a reduced main dump), which fits naturally into the planned follow-up issue on unifying the two particle write paths; I'll note it there.

HsunYeong:
@barrytchiang 
- I'll prefer everything in a single file (maybe called `SubData_*`) which contains both grid and particle data, and make both massive particle and tracer particle in the `Particle` group (if both output options are on).
- For output precision, I think making a separate option to control both particle types might be more user friendly.

barrytchiang:
@HsunYeong Thanks! I agree on the precision point: I'll replace the two 0/1/2 int flags with boolean `OPT__OUTPUT_SUBDIV_PAR`/`OPT__OUTPUT_SUBDIV_TRACER` plus a single shared `OPT__OUTPUT_SUBDIV_FLOAT32` controlling both particle types.

On the single combined `SubData_*` file: I like it and I'm happy to implement it, but I'd suggest offering it as one of two modes rather than the only one, since the two serve genuinely different use cases and the size difference is significant for the compact one.

Some measured numbers from my cold stream-disk 3-pc production runs (620k-patch hierarchy, 1.1e7 particles of which 1e6 are tracers, tracer sub-dumps at 5x the main-dump cadence):
- main dump `Data_*`: 32.4 GB = 30.6 GB GridData + 1.57 GB Particle + 0.185 GB Tree
- tracer-only compact sub-dump: 0.08-0.31 GB (float32)

A main-dump-formatted `SubData_*` must carry the `Tree` metadata for the patch-ordered layout (and for yt), so even a tracers-only sub-dump grows by ~0.19 GB here: a 1.6-3x increase per file, roughly doubling the total sub-dump footprint of a long run. More generally, the compact format scales as O(N_selected particles) while the combined format scales as O(N_patches + N_particles), so for patch-rich, tracer-sparse runs the metadata dominates the file. And this metadata buys nothing for workflows that never visualize sub-dumps with yt in the first place: high-cadence tracer sub-dumps are often consumed purely by scripted Lagrangian analysis (PUID-matched attribute reads), where the flat per-attribute datasets are already all that is needed.

Proposal: keep the current compact flat files as the default behavior of `OPT__OUTPUT_SUBDIV_PAR`/`_TRACER`, and add the combined mode as an option (e.g., `OPT__OUTPUT_SUBDIV_COMBINE`: write a single `SubData_*` in the main-dump format containing the `Input__Sub_Grid`-selected grid fields plus a `Particle` group with the enabled particle types, fully yt-loadable, with `OPT__OUTPUT_SUBDIV_FLOAT32` applying to its particle data). This lets users pick the mode that fits their science: one yt script for both main dumps and sub-dumps (directly addressing @hyschive's point) when visualization matters, or minimum-marginal-cost output when cadence and storage are the constraint. Implementation-wise the combined mode is mostly a call into the existing main-dump writer with the sub-dump field mask, and the compact writer already shares the HDF5 helpers after the (a) relocation, so the added maintenance surface is essentially one boolean plus documentation. Would that work for you?

That said, if you and @hyschive still feel a single uniform format is the better long-term design for the code, that's a call I'm happy to defer to -- I'll implement the combined `SubData_*` as the only mode and adjust accordingly.

HsunYeong:
@barrytchiang Thanks for the explanation, I think your proposed method can work, which can potentially be integrated into `OPT__OUTPUT_PAR_MODE`.
@hyschive Any suggestions?

hyschive:
@barrytchiang @HsunYeong 

### Major suggestions
- I would prefer to _always use a single-output-file format_ rather than having two different modes, so that the same `yt` script can be applied to both full and sub data.
- To address the concern about the extra storage required by the tree structure, I suggest further adding options to _output the tree, grids, and particles independently_, corresponding to the `Tree`, `GridData`, and `Particle` groups in the HDF5 output.

### More details
- I don’t think the following comment is quite correct. The data for each particle attribute are still stored as a single continuous array. The difference is that they are patch-ordered and cannot be loaded by `yt` unless the tree structure is also recorded.
> A main-dump-formatted SubData_* must carry the Tree metadata for the patch-ordered layout
- If data size is really a concern, one can output only the particle data without the tree or grids. The resulting data size should be very close to that of your current approach.
- In short, the major differences in the particle data format between the proposed and current approaches are:
  - The former is patch-ordered, while the latter is not. Both store each particle attribute as a single contiguous array and therefore consume almost the same amount of disk space for a given set of particles. Thus, the following advantage should still apply to the proposed new approach.
  > SubParticle_*/SubTracer_* are compact, flat, self-describing HDF5 files intended for direct h5py/NumPy access (each attribute is a named 1D dataset), which is what makes the high-cadence tracer analysis cheap
  - The former mixes all particle types in a single array, while the latter separates tracer and massive particles.
    - I don’t think this is a serious issue, since filtering by particle type can be done straightforwardly and inexpensively in post-processing.
    - This is mainly to conform to the current data format. We should definitely improve this for both full and sub dumps. See, for example, the discussion [here](https://github.com/gamer-project/gamer/issues/539#issuecomment-5267682028).
  - The former stores all grid and particle data in a single file, just like a full snapshot.
- When outputting only a subset of particle types, we may need to overwrite particle-count-related variables in the HDF5 file for `yt` to work correctly.
- Output the tree by default.
- When outputting grid data, the tree must also be output.

### Other minor comments
> I’ll keep the absent-Input__Sub_Grid fallback mapping to the OPT__OUTPUT_* flags (i.e., ( SubGridMode && !SubGridDerAll ) ? ... : ...), since with the file absent SubGrid_DerFieldSelected() returns true for every label and sub-dumps would otherwise write all derived fields regardless of the flags.
- Why not simply raise an error when `Input__Sub_Grid` is missing? It will keep `OPT__OUTPUT_*` and `Input__Sub_Grid` separate, which seems clearer to me.
- I just noticed that you and Hsun-Yeong have already agreed on this point.

> I’ll document: ParDens/TotalDens still requires OPT__OUTPUT_PAR_DENS (the flag selects the deposition mode), and user-defined derived fields still require OPT__OUTPUT_USER_FIELD (their labels are only initialized when it is enabled).
- Raise an error when these requirements are not satisfied..
- Does it mean that if `ParDens / TotalDens` is stored in sub dumps, it must also be stored in full dumps? It’s probably acceptable for now, though a bit counterintuitive. We should remove such dependency in a follow-up PR. Please add it to the to-do issue..

Apologies for the late comments and the extra work they cause! Any alternative suggestions are welcome. If helpful, we can have an online meeting sometime this week to clarify everything.

barrytchiang:
@hyschive @HsunYeong Thanks for the detailed suggestions! And thanks for the correction on the particle layout: you're right that patch ordering does not by itself require storing the tree (it is needed only for yt to interpret the ordering). With the independent group toggles, the particles-only configuration is then essentially identical in size to the current compact files, which fully resolves the storage concern.

I've adopted the single-format design accordingly. The latest commits implement and validate it, and the PR description is fully updated with the complete specification; a brief summary:
- A single `SubData_%06d` file (indexed by `SubDumpID`) in the full-snapshot HDF5 format replaces the previous `SubGrid_*`/`SubParticle_*`/`SubTracer_*` files; `Par_Output_SubDump.cpp` is removed and sub-dump output is unified into `Output_DumpData_Total_HDF5()`. This also supersedes the earlier two-mode proposal (and folding the mode selection into `OPT__OUTPUT_PAR_MODE`).
- Independent group toggles: `_TREE` [1] (`Tree`), `_GRID` [0] (`GridData`, field list from the now-required `Input__Sub_Grid`), and `_PAR`/`_TRACER` [0] (massive/tracer particles, mixed in one patch-ordered array as in full dumps; separated via `ParType` in post-processing, with the broader type-separation improvement added to the to-do issue per the #539 discussion). `_FLOAT32` [0] downcasts the particle data to single precision, replacing the former per-type `= 2` modes; the former tracer `ParMass` suppression is dropped for format conformity. Storage-critical high-cadence particle output is `_TRACER=1`, `_TREE=0`, `_FLOAT32=1`; the defaults keep every `SubData_*` yt-loadable out of the box.
- When only a subset of particle types is output, the particle counts in `KeyInfo` and `Tree/NPar` are rewritten to the type-filtered counts so that yt indexes the `Particle` arrays correctly (verified explicitly with the ParticleTest problem).
- Per your comments, a missing `Input__Sub_Grid` and the `ParDens`/user-derived-field flag violations are now fatal errors rather than documentation-only.

Validated end to end (details in the PR description): the full (Plummer, including the new `plot_subdata.py` yt script), minimal (`Info`/`Particle`/`User` + float32), and type-filtered (ParticleTest) configurations all load through the standard yt gamer frontend; the Zeldovich `_USER` path works unchanged, including a pre-existing `Input__Parameter` running correctly on the new defaults; and the compile matrix covers HYDRO/ELBDM/MHD/SRHD/GRACKLE.

doc/wiki/Simulation-Logs-related/[Simulation-Logs]-Record__TimeSubDump.md
```diff
@@ -0,0 +1,23 @@
...
+each sub-cadence output (see
+[[OPT__OUTPUT_SUBDIV | [Runtime-Parameters]-Outputs#OPT__OUTPUT_SUBDIV]]),
+including the sub-dump events coinciding with the main data dumps.
+
+Example:
+``` markdown
+SubDumpID                       Time                 Step
+        0          0.00000000000000e+00                    0
+        1          2.50000000000000e-01                    5
+        2          5.00000000000000e-01                   10
```
Comment (HsunYeong) [inline doc/wiki/Simulation-Logs-related/[Simulation-Logs]-Record__TimeSubDump.md]:
```suggestion
SubDumpID                       Time                 Step
        0       0.00000000000000e+00                    0
        1       2.50000000000000e-01                    5
        2       5.00000000000000e-01                   10
```

hyschive:
@barrytchiang Thanks for the update!

@HsunYeong Given this non-trivial refactor, could you review the PR again? Thanks!

src/Output/Output_L1Error.cpp
```diff
@@ -57,7 +57,10 @@ void Output_L1Error( void (*AnalFunc_Flu)( real fluid[], const double x, const d
                      const char *Prefix, const OptOutputPart_t Part, const double x, const double y, const double z )
 {
 
-   if ( MPI_Rank == 0 )    Aux_Message( stdout, "%s (DumpID = %d) ...\n", __FUNCTION__, DumpID );
+// use SubDumpID for file naming when OPT__OUTPUT_SUBDIV is enabled so sub-dump files don't collide with main dumps
+   const int FileID = ( OPT__OUTPUT_SUBDIV >= 1 ) ? SubDumpID : DumpID;
+
+   if ( MPI_Rank == 0 )    Aux_Message( stdout, "%s (DumpID = %d, SubDumpID = %d) ...\n", __FUNCTION__, DumpID, SubDumpID );
```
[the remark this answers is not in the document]
barrytchiang: Good catch, agreed. Fixed in ac52a355c: `FileID` uses `SubDumpID` only when `OPT__OUTPUT_SUBDIV_USER` is enabled.

src/Init/Init_ByRestart_HDF5.cpp
```diff
@@ -206,6 +206,7 @@ void Init_ByRestart_HDF5( const char *FileName )
    MPI_Barrier( MPI_COMM_WORLD );
 
    LoadField( "DumpID",               &KeyInfo.DumpID,               H5_SetID_KeyInfo, H5_TypeID_KeyInfo,    Fatal,  NullPtr,              -1, NonFatal );
+   LoadField( "SubDumpID",            &KeyInfo.SubDumpID,            H5_SetID_KeyInfo, H5_TypeID_KeyInfo, NonFatal,  NullPtr,              -1, NonFatal );
    LoadField( "NX0",                   KeyInfo.NX0,                  H5_SetID_KeyInfo, H5_TypeID_KeyInfo,    Fatal,  NX0_TOT,               3,    Fatal );
```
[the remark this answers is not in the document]
barrytchiang: Agreed, `LoadField()` indeed leaves the field untouched when absent. Fixed in ac52a355c: `KeyInfo.SubDumpID` now defaults to `-1` and restores as `0` for pre-feature snapshots.

src/Init/Init_ByRestart_HDF5.cpp
```diff
@@ -2468,6 +2475,10 @@ void Check_InputPara( const char *FileName, const int FormatVersion )
    LoadField( "Output_PartZ",                &RS.Output_PartZ,                SID, TID, NonFatal, &RT.Output_PartZ,                1, NonFatal );
    }
    LoadField( "InitDumpID",                  &RS.InitDumpID,                  SID, TID, NonFatal, &RT.InitDumpID,                  1, NonFatal );
+   LoadField( "Opt__Output_Subdiv_Grid",     &RS.Opt__Output_Subdiv_Grid,     SID, TID, NonFatal, &RT.Opt__Output_Subdiv_Grid,     1, NonFatal );
+   LoadField( "Opt__Output_Subdiv_Par",      &RS.Opt__Output_Subdiv_Par,      SID, TID, NonFatal, &RT.Opt__Output_Subdiv_Par,      1, NonFatal );
+   LoadField( "Opt__Output_Subdiv_Tracer",   &RS.Opt__Output_Subdiv_Tracer,   SID, TID, NonFatal, &RT.Opt__Output_Subdiv_Tracer,   1, NonFatal );
+   LoadField( "Opt__Output_Subdiv_User",     &RS.Opt__Output_Subdiv_User,     SID, TID, NonFatal, &RT.Opt__Output_Subdiv_User,     1, NonFatal );
```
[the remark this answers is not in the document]
barrytchiang: Added the `Opt__Output_Subdiv`, `_Tree`, and `_Float32` rows in ac52a355c.

src/Init/Init_ByRestart_HDF5.cpp
```diff
@@ -358,6 +359,12 @@ void Init_ByRestart_HDF5( const char *FileName )
    else
       DumpID = INIT_DUMPID;
 
+// 1-8b. recover user sub-dump state
+   if ( ! OPT__RESTART_RESET )
+   {
+      SubDumpID = KeyInfo.SubDumpID + 1;
+   }
```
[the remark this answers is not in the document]
barrytchiang: You're right, and this was subtle: the `+1` dated from an earlier ordering, but the unified writer now increments `SubDumpID` before writing the co-dump snapshot, so the stored value is already the next-to-use ID and `+1` skipped one per restart. Fixed in ac52a355c to restore directly (with a `-1` to `0` fallback for pre-feature snapshots) and verified with an explicit restart test: restarting from a snapshot storing `SubDumpID = 3` resumes writing `SubData_000003` with no gap or collision.

example/test_problem/Hydro/Plummer/Input__Parameter
```diff
@@ -211,6 +211,20 @@ OPT__OUTPUT_LORENTZ           0           # output Lorentz factor [0] ##SRHD ONL
 OPT__OUTPUT_3VELOCITY         0           # output 3-velocities [0] ##SRHD ONLY##
 OPT__OUTPUT_USER_FIELD        0           # output user-defined derived fields [0] -> edit "Flu_DerivedField_User.cpp"
 OPT__OUTPUT_MODE              2           # (1=const step, 2=const dt, 3=dump table) -> edit "Input__DumpTable" for 3
+OPT__OUTPUT_SUBDIV           -1           # sub-cadence gate: <0=off; N>=1: sub-dumps between/at main dumps [-1]
```
HsunYeong: Should we make it turned on by default? Since setting `OPT__OUTPUT_SUBDIV_GRID=1` with this option off will trigger error in `Aux_Check_Parameter()`. Or we can remove the `if ( AnySubDiv  &&  OPT__OUTPUT_SUBDIV < 1 )` check.

src/Output/Output_L1Error.cpp
```diff
@@ -57,7 +57,11 @@ void Output_L1Error( void (*AnalFunc_Flu)( real fluid[], const double x, const d
                      const char *Prefix, const OptOutputPart_t Part, const double x, const double y, const double z )
 {
 
-   if ( MPI_Rank == 0 )    Aux_Message( stdout, "%s (DumpID = %d) ...\n", __FUNCTION__, DumpID );
+// use SubDumpID for file naming only when user outputs fire at sub-cadence (OPT__OUTPUT_SUBDIV_USER)
+// so sub-dump files don't collide with main dumps
+   const int FileID = ( OPT__OUTPUT_SUBDIV_USER  &&  OPT__OUTPUT_SUBDIV >= 1 ) ? SubDumpID : DumpID;
+
+   if ( MPI_Rank == 0 )    Aux_Message( stdout, "%s (DumpID = %d, SubDumpID = %d) ...\n", __FUNCTION__, DumpID, SubDumpID );
```
HsunYeong: print `DumpID` only when SubDump is disabled.

src/Output/Output_L1Error.cpp
```diff
@@ -340,7 +344,7 @@ void Output_L1Error( void (*AnalFunc_Flu)( real fluid[], const double x, const d
    } // if ( MPI_Rank == 0 )
 
 
-   if ( MPI_Rank == 0 )    Aux_Message( stdout, "%s (DumpID = %d) ... done\n", __FUNCTION__, DumpID );
+   if ( MPI_Rank == 0 )    Aux_Message( stdout, "%s (DumpID = %d, SubDumpID = %d) ... done\n", __FUNCTION__, DumpID, SubDumpID );
```
HsunYeong: print DumpID only when SubDump is disabled.

HsunYeong (commented):
@barrytchiang Thanks for the update! Minor comments added.

example/test_problem/Hydro/Plummer/Input__Parameter
barrytchiang: Thank you for catching this bug in the example! With the shipped presets `OPT__OUTPUT_SUBDIV_GRID/PAR = 1` and the gate off, Plummer aborted in `Aux_Check_Parameter()` out of the box. Fixed in 7d62d26d3 by turning the gate on `(OPT__OUTPUT_SUBDIV = 2)` rather than dropping the check, so the example runs as shipped and demonstrates the sub-cadence output (matching the new README section and `plot_script/plot_subdata.py`), while the check still catches enabling a sub-flag while forgetting the gate, which would otherwise silently produce no sub-dumps. Happy to switch if you'd rather Plummer keep the upstream default of writing no extra files.

src/Output/Output_L1Error.cpp
barrytchiang: Done in 7d62d26d3: the message now prints `SubDumpID` when the sub-cadence path names the files and `DumpID` otherwise, instead of always printing both.

PR #537, opened 2026-08-02
Allow the initial simulation time to be set for the initial conditions

hsinhaoHHuang:
## Goals
- Provide more flexibility for setting the initial simulation time. Previously, it was fixed to `0.0` for non-comoving simulations.

## Changes
- Added a new runtime parameter `INIT_TIME` as the initial `Time[lv]` the simulation starts from.
  - For non-comoving simulations, the default is `0.0`, and it can be set to any non-negative value.
  - For comoving simulations, the value is always overwritten by the existing parameter `A_INIT` (the initial scale factor).

- Added validation requiring the ending time to be no earlier than the initial time.
- Simulation notes and HDF5 outputs now record the initial time.
- Documented defaults, limits, behavior, and the relationship between INIT_TIME and A_INIT

## Notes
- Currently, the allowed lower bound of `INIT_TIME` assumes `Time` is `>= 0.0`.

## Tests Done
The following test problems are tested with their (1) fresh runs of default and changed `INIT_TIME` and (2) restart runs.
- `Hydro/Plummer`
- `Hydro/CDM_LSS`
- `ELBDM/JeansInstabilityComoving` 

All behave as expected.

## Poem by CodeRabbit

> A rabbit checks the starting time,
> Zero-point bounds in neat design.
> Through restart paths the clocks align,
> HDF5 keeps the value fine.
> In comoving runs, A_INIT shines.


src/Init/Init_ByRestart_HDF5.cpp
```diff
@@ -330,7 +330,7 @@ void Init_ByRestart_HDF5( const char *FileName )
 // --> assuming dTime_AllLv[] has been initialized as 0.0 properly
    for (int lv=KeyInfo.NLevel; lv<NLEVEL; lv++)
    {
-      Time              [lv] = 0.0;
+      Time              [lv] = INIT_TIME;
```
hsinhaoHHuang, line 333: Should these extra levels be set with the same "restart" time?

```suggestion
      Time              [lv] = ( OPT__RESTART_RESET ) ? INIT_TIME : Time[0];
```

src/Init/Init_Load_Parameter.cpp
```diff
@@ -37,6 +37,7 @@ void Init_Load_Parameter()
    ReadPara->Add( "MPI_NRANK_Z",                &MPI_NRank_X[2],                 -1,               NoMin_int,     NoMax_int      );
 // do not check OMP_NTHREAD since it may be reset by Init_ResetParameter()
    ReadPara->Add( "OMP_NTHREAD",                &OMP_NTHREAD,                    -1,               NoMin_int,     NoMax_int      );
+   ReadPara->Add( "INIT_TIME",                  &INIT_TIME,                       0.0,             0.0,           NoMax_double   );
```
hsinhaoHHuang, line 40: Will it be fine to allow a negative `Time`?

PR #543, opened 2026-08-14
Updates for ClusterMerger and new NASA machine configs

jzuhone:
This PR is a collection of two updates for the ClusterMerger problem and updating configs for the machines at NASA's Ames Research Center.

## Improving the transition from cluster profiles to background

The way to set up the gas properties for the clusters for the `ClusterMerger` problem involves interpolating density and pressure from radial profiles onto the grid cells and adding them from the different clusters. Currently, the code handles the boundaries of the clusters by taking the value of density/pressure at the largest radius of the profile and allowing this to be the extrapolation, which means at large radii that the total density/pressure will be the individual sum of each, but that at radii where one cluster is still within its profile and the other is not there will still be a varying profile. If we are within the maximum radius for a cluster at a grid point, the momentum density vector from that cluster is added as well.

This seems sensible in practice, but the result often is that one gets discontinuous features at the boundaries of the clusters in the initial condition itself in the temperature and velocity, like seen in the figure below:

Data_000000_Slice_z_kT

Data_000000_Slice_z_velocity_x

which we have seen has adverse effects in simulations, as the artifacts have downstream effects later, see:

Data_000030_Slice_z_kT

Data_000030_Slice_z_velocity_x

This PR handles this problem by introducing two new parameters to `Input__TestProb` that impose a background density and temperature (the latter giving a background pressure when combined with the former), and the density and temperature cannot fall below the specified values outside of the clusters in the initial condition. The transition from the boundary of the cluster to the background is continuous everywhere. We allow the momenta to be set initially based on the profiles everywhere in the domain instead of cutting it off at the maximum radius of the profile, which allows for a continuous transition of the momentum and velocity to a background value (because the momentum drops while the density stays flat in the background, the asymptotic value of the velocity in the background is zero. The result is the much better initial condition without the artifacts:

Data_000000_Slice_z_kT2

Data_000000_Slice_z_velocity_x2

## Allowing the z-position and z-velocity of the clusters to be specified by the user

This PR also allows the user to specify the z-position and z-velocity of the cluster centers--previously the default was to keep them at the domain center and zero. This is still the default if the corresponding parameters are not specified.

## Updating machine configurations for the NASA Ames machines. 

The Pleiades machine at NASA Ames has been decommissioned, and the new GPU machine is called Cabeus, and has two types of CPU/GPU nodes, one with x86 arch and the other with arm. This PR removes the old pleiades machine file and adds two more for cabeus.


jzuhone:
@hyschive I've resolved all the comments suggested by Copilot after testing.

hyschive:
@jzuhone I’m currently at a workshop and will review this PR ASAP.

src/TestProblem/Hydro/ClusterMerger/Init_TestProb_ClusterMerger.cpp
```diff
@@ -761,6 +774,9 @@ void SetGridIC( real fluid[], const double x, const double y, const double z, co
          else              fluid[ColorFieldsIdx[c]] = 0.0;
       }
    } // for (int c=0; c<Merger_Coll_NumHalos; c++)
+   
+   Dens = MAX( Dens, Merger_Coll_BkgDensity );
+   Pres = MAX( Pres, Dens*Merger_Coll_BkgTemperature );
```
hyschive, line 779: How about using the EoS routine `EoS_DensTemp2Pres_CPUPtr()` to improve clarity and extensibility? If you agree, please remove line 427 accordingly.

src/TestProblem/Hydro/ClusterMerger/Init_TestProb_ClusterMerger.cpp
hyschive, line 1: Please remove the trailing spaces around line 777.

src/TestProblem/Hydro/ClusterMerger/Init_TestProb_ClusterMerger.cpp
```diff
@@ -295,30 +299,31 @@ void LoadInputTestProb( const LoadParaMode_t load_mode, ReadPara_t *ReadPara, HD
...
       if ( AGN_feedback ) {
       LOAD_PARA( load_mode, CM_BH_Mass_name,           &CM_BH_Mass[c],           -1.0,                Eps_double,    NoMax_double   );
       LOAD_PARA( load_mode, Jet_HalfHeight_name,       &Jet_HalfHeight[c],       -1.0,                Eps_double,    NoMax_double   );
       LOAD_PARA( load_mode, Jet_Radius_name,           &Jet_Radius[c],           -1.0,                Eps_double,    NoMax_double   );
       }
-   } // for (int c=0; c<Merger_Coll_NumHalos; c++)
-
-      LOAD_PARA( load_mode, "Merger_Coll_UseMetals",   &Merger_Coll_UseMetals,    true,               Useless_bool,  Useless_bool   );
-
+   } // for ( int c=0; c<Merger_Coll_NumHalos; c++ )
```
hyschive:
```suggestion
   } // for (int c=0; c<Merger_Coll_NumHalos; c++)
```

src/TestProblem/Hydro/ClusterMerger/Init_TestProb_ClusterMerger.cpp
```diff
@@ -12,6 +12,8 @@ static char   (*Merger_File_Prof)[ MAX_STRING ] = NULL; // profile table of clus
 static bool    *Merger_Coll_IsGas = NULL;               // (true/false) --> does cluster have gas
        double (*Merger_Coll_Pos)[3] = NULL;             // initial position of clusters
        double (*Merger_Coll_Vel)[3] = NULL;             // initial velocity of clusters
+       double   Merger_Coll_BkgDensity;                 // background gas density
+       double   Merger_Coll_BkgTemperature;             // background gas temperature
```
hyschive, line 16: I would recommend adding these two parameters to `example/test_problem/Hydro/ClusterMerger/Input_Options/default/Input__TestProb` to make them more visible. Also, please clarify their respective units and default values there.

src/TestProblem/Hydro/ClusterMerger/Init_TestProb_ClusterMerger.cpp
```diff
@@ -295,30 +299,31 @@ void LoadInputTestProb( const LoadParaMode_t load_mode, ReadPara_t *ReadPara, HD
...
       sprintf( Jet_Radius_name,        "Jet_Radius%d",        c+1 );
       }
-
       LOAD_PARA( load_mode, Merger_File_Prof_name,      Merger_File_Prof[c],      NoDef_str,          Useless_str,   Useless_str    );
       LOAD_PARA( load_mode, Merger_File_Par_name,       Merger_File_Par[c],       NoDef_str,          Useless_str,   Useless_str    );
       LOAD_PARA( load_mode, Merger_Coll_IsGas_name,    &Merger_Coll_IsGas[c],     true,               Useless_bool,  Useless_bool   );
       LOAD_PARA( load_mode, Merger_Coll_PosX_name,     &Merger_Coll_Pos[c][0],   -1.0,                NoMin_double,  NoMax_double   );
       LOAD_PARA( load_mode, Merger_Coll_PosY_name,     &Merger_Coll_Pos[c][1],   -1.0,                NoMin_double,  NoMax_double   );
-      LOAD_PARA( load_mode, Merger_Coll_VelX_name,     &Merger_Coll_Vel[c][0],   -1.0,                NoMin_double,  NoMax_double   );
-      LOAD_PARA( load_mode, Merger_Coll_VelY_name,     &Merger_Coll_Vel[c][1],   -1.0,                NoMin_double,  NoMax_double   );
```
hyschive, line 312: [Bug] Please restore these two lines.

src/TestProblem/Hydro/ClusterMerger/Init_TestProb_ClusterMerger.cpp
```diff
@@ -284,8 +286,10 @@ void LoadInputTestProb( const LoadParaMode_t load_mode, ReadPara_t *ReadPara, HD
       char Merger_Coll_IsGas_name[MAX_STRING];
       char Merger_Coll_PosX_name [MAX_STRING];
       char Merger_Coll_PosY_name [MAX_STRING];
+      char Merger_Coll_PosZ_name [MAX_STRING];
```
hyschive, line 289: Similarly, let's add it to `Input_Options/default/Input__TestProb`.

src/TestProblem/Hydro/ClusterMerger/Init_TestProb_ClusterMerger.cpp
```diff
@@ -284,8 +286,10 @@ void LoadInputTestProb( const LoadParaMode_t load_mode, ReadPara_t *ReadPara, HD
...
       char Merger_Coll_VelX_name [MAX_STRING];
       char Merger_Coll_VelY_name [MAX_STRING];
+      char Merger_Coll_VelZ_name [MAX_STRING];
```
hyschive, line 292: Similarly, let's add it to `Input_Options/default/Input__TestProb`.

src/TestProblem/Hydro/ClusterMerger/Init_TestProb_ClusterMerger.cpp
```diff
@@ -295,30 +299,31 @@ void LoadInputTestProb( const LoadParaMode_t load_mode, ReadPara_t *ReadPara, HD
...
+      sprintf( Merger_Coll_PosZ_name,  "Merger_Coll_PosZ%d",  c+1 );
       sprintf( Merger_Coll_VelX_name,  "Merger_Coll_VelX%d",  c+1 );
       sprintf( Merger_Coll_VelY_name,  "Merger_Coll_VelY%d",  c+1 );
+      sprintf( Merger_Coll_VelZ_name,  "Merger_Coll_VelZ%d",  c+1 );
       if ( AGN_feedback ) {
       sprintf( CM_BH_Mass_name,        "Bondi_MassBH%d",      c+1 );
       sprintf( Jet_HalfHeight_name,    "Jet_HalfHeight%d",    c+1 );
       sprintf( Jet_Radius_name,        "Jet_Radius%d",        c+1 );
       }
-
```
hyschive, line 305: How about restoring this empty line to better separate `sprintf()` and `LOAD_PARA()`?

src/TestProblem/Hydro/ClusterMerger/Init_TestProb_ClusterMerger.cpp
hyschive, line 312: To be more cautious, how about setting the default values of `Merger_Coll_Pos[c][0/1]` and `Merger_Coll_Vel[c][0/1]` to, for example, `2.0*NoMax_double`, so that they trigger an error if any of them are not set in `Input__TestProb`?

src/TestProblem/Hydro/ClusterMerger/Init_TestProb_ClusterMerger.cpp
```diff
@@ -404,11 +409,22 @@ void SetParameter()
...
-      Merger_Coll_Pos[c][1] *= Const_kpc / UNIT_L;
-      Merger_Coll_Vel[c][0] *= (Const_km/Const_s) / UNIT_V;
-      Merger_Coll_Vel[c][1] *= (Const_km/Const_s) / UNIT_V;
+
+      for ( int i = 0; i < 3; i++ )
+      {
+         Merger_Coll_Pos[c][i] *= Const_kpc / UNIT_L;
+         Merger_Coll_Vel[c][i] *= (Const_km/Const_s) / UNIT_V;
+      }
+      // Negative values for the z-axis position indicate a merger in x-y plane
```
hyschive, line 418: How about also adding this comment to `Input_Options/default/Input__TestProb` for clarification?

src/TestProblem/Hydro/ClusterMerger/Init_TestProb_ClusterMerger.cpp
```diff
@@ -404,11 +409,22 @@ void SetParameter()
...
+      if ( Merger_Coll_Pos[c][2] < 0.0 )
+      {
+         Merger_Coll_Pos[c][2] = amr->BoxCenter[2];
+         Merger_Coll_Vel[c][2] = 0.0;
+      }
+
    }
+   Merger_Coll_BkgDensity /= UNIT_D;
+   Merger_Coll_BkgTemperature *= Const_kB * UNIT_M / ( UNIT_E * MOLECULAR_WEIGHT * MU_NORM );
```
hyschive, line 427: See my comment at line 779.

src/TestProblem/Hydro/ClusterMerger/Init_TestProb_ClusterMerger.cpp
```diff
@@ -771,8 +787,10 @@ void SetGridIC( real fluid[], const double x, const double y, const double z, co
    fluid[MOMY     ] = MomY;
    fluid[MOMZ     ] = MomZ;
    fluid[ENGY     ] = Etot;
+
+   real Z_sol = 0.0165; // approximate value of solar metallicity as mass fraction
    if ( Merger_Coll_UseMetals )
-   fluid[Idx_Metal] = Metl;
+   fluid[Idx_Metal] = MAX( Metl, 0.3*Z_sol*Dens );
```
hyschive, line 793: How about adding a brief comment explaining this literal constant 0.3?

hyschive (changes requested):
@jzuhone Thanks for the PR! I've added some comments. They are all minor except for one bug fix.

src/TestProblem/Hydro/ClusterMerger/Init_TestProb_ClusterMerger.cpp
jzuhone: I would rather skip this, since the default value is `-1.0`, if the Z-position is not specified, and I don't expect that any end-user would set this explicitly.

jzuhone:
@hyschive ok, addressed these, thanks!

hyschive (approved):

PR #551, opened 2026-09-14
Add stricter total energy and internal energy checks in Flu_Close()

hsinhaoHHuang:
## Issue
- With the changes introduced by #504, when the gas has `Eint < -Etot`, we deliberately skip the dual-energy fix in `Hydro_FullStepUpdate()`. We hope the negative `Etot` or `Eint` can be caught in `Flu_Close()` and then trigger subsequent corrections.
However, in `Flu_Close()`, we don’t check for either negative `Etot` or `Eint` in the dual-energy formalism. As long as all the other variables are normal, the unphysical `Etot` and `Eint` remain undetected.
For example, the following three cases can possibly be found in the stage of `Flu_Close()` and **DO NOT** return unphysical:
   1. Negative total energy
      `[Dens =  1.0, MomX =  1.0, MomY =  0.0, MomZ =  0.0, Engy = -0.1, Dual =  1.0]`
      --> `[Ekin =  0.5, Eint = -0.6, Etot = -0.1]`
   2. Large negative internal energy due to the skipped dual-energy fix
      `[Dens =  1.0, MomX =  1.0, MomY =  0.0, MomZ =  0.0, Engy =  0.2, Dual =  1.0]`
      --> `[Ekin =  0.5, Eint = -0.3, Etot = 0.2]`
   3. Small negative internal energy due to the rounding error
      `[Dens =  1.0, MomX =  1.0, MomY =  0.0, MomZ =  0.0, Engy =  0.5-epsilon, Dual =  1.0]`
      --> `[Ekin =  0.5, Eint = -epsilon, Etot = 0.5-epsilon]`

   While the third case here is acceptable, the first two cases should return unphysical.
- Stricter unphysical checks are preferred in general to avoid further problems.

## Changes
- Always check `Fluid[ENGY]` against `MIN_EINT` in `Unphysical()` in `Flu_Close()`
   - `Fluid[ENGY]` should not be less than `MIN_EINT` anyway. And this check is cheap.
- Change the argument `int CheckMode` to `bool CheckMinEint` as `Etot` is always checked now.
- When `DUAL_ENERGY` is enabled, check `Eint` against the negative rounding-error scale of `Etot`.
- When `DUAL_ENERGY` is enabled, always check the pressure converted from `DUAL` and `DUAL` itself, even if `AutoReduceDt_Continue` is false.
- Check `Eint` against the negative rounding-error scale of Etot, even if `AutoReduceDt_Continue` is false.

## Tests
Run the test problem `Hydro/AGORA_IsolatedGalaxy` with the low-resolution setup:

|        | Original | Check Etot only | Check Eint only |
|--------|--------|--------|--------|
| number of undetected negative total energies in `Flu_Close()` printed during the first step | 41  | 0 | 0 |
| `NCorrAllLv` during the first step in `Record__NCorrUnphy` | 0 | 16 | 22 |
| Density Projections at t = 300 Myr | image | image | image | 
| Temperature-Density Diagram at t = 300 Myr | image | image | image | 
| Star Formation Rate | image | image | image | 

- `OPT__1ST_FLUX_CORR` is enabled as defualt
- In the final implementation, we check both.


barrytchiang (commented):
@hsinhaoHHuang thanks for preparing this PR, and your case 1 is exactly what we hit in production. Our disk+stream runs (`EOS_GAMMA`, `DUAL_ENERGY=DE_ENPY`) also show cells with negative `Etot` and a healthy `Dual`, up to 75 cells persisting across seven consecutive dumps. I think it's worth keeping both of the new tests rather than only the cheap one. Since the dual-energy fix in `Hydro_FullStepUpdate()` is skipped whenever `Etot*2 <= Ekin`, which includes cells with positive `Etot`, a test on `Etot` alone would still let your case 2 through.

**One potential issue: in the final check, the new floor comparisons can abort the run.** When `AutoReduceDt_Continue` is false, `CorrectUnphysical()` first applies the floors enabled by `OPT__LAST_RESORT_FLOOR`, so `Hydro_DualEnergyFix()` sets `Pres = MIN_PRES` and `Dual = Hydro_DensPres2Dual(Dens, MIN_PRES)`. `Unphysical()` then recomputes `Pres = Hydro_DensDual2Pres(Dens, Dual)` and tests `Pres < MIN_PRES` exactly. That round trip, `POW(Dens,-Gamma_m1)` followed by `POW(Dens,Gamma_m1)`, is not exact, so a cell the floor has just repaired can come back one rounding step below the floor and fail, where `main` accepted it (the old `CheckMinEtot` mode did not test the dual pressure).

These failure cases are also reproducible in the `Riemann` test problem, with `Riemann_Prob 3` (Einfeldt 1-2-0-3) with `MIN_PRES 0.02`: `main` runs to t=0.1, while this branch stops with `Aux_Error` "fluid solver failed" at step 8 in double precision and step 7 in single precision, and also with the default `AUTO_REDUCE_DT 1`, where the step is retried with smaller dt 161 times before the run stops at the same point. In the unphysical cell the diagnostic says `total energy density update = MinPres` and every other new test passes with a wide margin (`Etot` 3.32e-2 > `MIN_EINT` 3.0e-2, `Etot-Ekin` = 3.0e-2, `Dual` 0.159), which leaves the exact `Pres < MIN_PRES` comparison as the only condition that can be failing here.

My suggestion is to keep the new `Etot` and `Eint` tests as they are, and make only the floor comparisons rounding aware in the final check, when `AutoReduceDt_Continue` is false, e.g. compare against `MIN_PRES*( 1 - CHECK_UNPHY_ROUNDING_FACTOR*MACHINE_EPSILON )` (and `TINY_NUMBER` rather than `2*TINY_NUMBER` for `Dual`), so a cell that the floor has just repaired cannot fail the check that follows it.

**One question.** The cells this PR now catches are repaired by the `Hydro_DualEnergyFix()` call in `CorrectUnphysical()`, which has no `Etot*2 > Ekin` guard, i.e. the case #504 deliberately avoids, and its comment (d) warns the resulting `Etot'` can be unphysically large when the density is tiny and the momentum finite. Is the 1st-order flux correction expected to remove that case before the fix runs, or should the same guard apply there?

**Minor.** Note 3 in the `Unphysical()` header still says that with the dual-energy formalism it checks pressure "instead of internal energy"; it now checks both.

hsinhaoHHuang:
@barrytchiang Thank you for the detailed review and great suggestions! I have updated the check to accept the rounding errors when comparing the pressure to `MIN_PRES` when `AutoReduceDt_Continue` is false. Please help me check again. Thanks!

Regarding the `Etot*2 > Ekin` guard of `Hydro_DualEnergyFix()`, I think it makes sense to also apply it here. I am considering adding it in a separate pull request to keep it focused. Thank you for the suggestion.

barrytchiang (commented):
@hsinhaoHHuang thanks for the update! I rebuilt at 72d5df7a and reran the Riemann test problem, under both single and double precision. The following setups can now run without termination error:
- `--dual=NONE`, double and single
- `DUAL_ENERGY=DE_ENPY`, double, `MIN_PRES 0` (bit-identical to `main` in every field of every dump)
- `DUAL_ENERGY=DE_ENPY`, single, `AUTO_REDUCE_DT 0` and 1 (correcting exactly as many cells as `main`, 4,416,512)

**`DUAL_ENERGY=DE_ENPY` in double precision still aborts, and it is the new `Eint` test that rejects the cell.** With the same `Riemann_Prob 3` (Einfeldt 1-2-0-3) and `MIN_PRES 0.02`, `main` runs to t=0.1 while this branch stops with "fluid solver failed" at step 7. With the default `AUTO_REDUCE_DT 1` the same step is retried 23 times, each attempt with smaller `MINMOD_COEFF` and `INT_MONO_COEFF` (`AUTO_REDUCE_DT_FACTOR` is 1.0 by default, so dt itself is not reduced), and then the run stops at the same place. I rebuilt with a debug print inside `Unphysical()`, and every rejected cell fails the new `Eint` comparison with identical numbers: `Eint = 0.029999999984999994` against `MIN_EINT = 0.03`, a shortfall of `1.5e-11` where the tolerance `Etot*CHECK_UNPHY_ROUNDING_FACTOR*MACHINE_EPSILON` is `2.6e-17`. These are cells that `Hydro_DualEnergyFix()` has just set to `Pres = MIN_PRES` in the same call (`DE_UPDATED_BY_MIN_PRES`, printed as `total energy density update = MinPres`), which is the failure mode noted in my last comment: the check immediately follows the floor that repaired the cell, and then rejects it. What has changed is which condition does the rejecting.

**The cause is that the pressure floor and the energy floor are derived from different values of gamma.** `Init_ResetParameter.cpp:1123-1128` sets `MIN_EINT = MIN_PRES*1.5`, which assumes gamma is exactly 5/3 and never consults `GAMMA`, while the floor block of `Hydro_DualEnergyFix()` sets `Pres = MinPres` and `Eint = Pres*_Gamma_m1` using the run's actual `GAMMA`. The `Riemann` example sets `GAMMA 1.666666667` in its `Input__Parameter`, so the two differ by a relative 5e-10. I have checked that this explanation reproduces the numbers bit for bit: `MIN_PRES*_Gamma_m1` gives exactly the `Eint` above, `Ekin + Eint` gives exactly the `Etot` in the same diagnostic, and the `Etot - Ekin` that `Unphysical()` recomputes returns exactly that `Eint` again. So there is no numerical round-off anywhere along the chain, and the mismatch comes entirely from the gamma inconsistency. On the same arithmetic, the floored `Eint` lies below `MIN_EINT` for any `GAMMA > 5/3` by a finite amount rather than a rounding scale (33% at `GAMMA = 2`), and the reverse branch `MIN_PRES = MIN_EINT/1.5` can have the same pathology.

`main` never hit this because its lenient pass checked `Etot` only. Here the `Eint` test is reached in every mode under `DUAL_ENERGY`, including the detection pass, so a bigger rounding tolerance will not help. One suggestion is to compare `Eint` against the internal energy that actually corresponds to `MIN_PRES` under the EoS in force, i.e. `EoS_DensPres2Eint(Dens, MIN_PRES)` (`MIN_PRES*_Gamma_m1` for `EOS_GAMMA`), rather than against `MIN_EINT`; or to make `MIN_EINT` consistent with `MIN_PRES` through the EoS at initialization, which would fit naturally in the separate PR you are already planning for the `Etot*2 > Ekin` guard.

**Two smaller things.** With `DUAL_ENERGY` enabled the `Eint` check always carries the rounding allowance, `Eint + Etot*CHECK_UNPHY_ROUNDING_FACTOR*MACHINE_EPSILON < MIN_EINT`, and never the bare `Eint < MIN_EINT`, including at the two `CheckStrictFloor_Yes` call sites. That is the right design and it is what catches your case 2, but the header note that used to describe the dual-energy behavior was removed, so nothing now says that a dual-energy build never checks `Eint` strictly, and the `CheckStrictFloor` parameter text reads as if it does. Also, `Hydro_Con2Eint()` is now evaluated for every cell even under `DUAL_ENERGY`, where it was previously skipped on purpose (the comment just above it still warns against computing pressure from `Etot-Ekin`); it is cheap for `EOS_GAMMA`, but it is the per-cell detection loop, so it may be worth a look.

